PR #19, opened 2026-05-01
Add Italian translation

albanobattistella:


maathimself:
Hey, thanks for putting this together — the translation looks great. Just two things blocking the merge:

1. Save the file as UTF-8

The file has encoding issues that are garbling all the Italian characters. In VS Code, click the encoding indicator in the bottom-right corner and choose "Save with Encoding → UTF-8". That should clean up everything like "giÃ " → "già", "â¦" → "…" etc.

2. Variables need double curly braces

i18next uses {{variable}} not {variable} — so things like {count} and {name} need to become {{count}} and {{name}} throughout. A global find/replace in your editor should do it in seconds.

Two tiny content fixes while you're in there:

admin.sso.errorRequired — missing "URL emittente" in the list
admin.messageList.threadingOnDesc — should be "Le risposte sono raggruppate in conversazioni"
Fix those up and push to the same branch and I'll get it merged!

maathimself:
Thank you for this, I pushing the changes myself. It is available in the source, will be in the next release.

albanobattistella:
> Thank you for this, I pushing the changes myself. It is available in the source, will be in the next release.

Hi,
Thanks to you!

PR #28, opened 2026-05-07
fix: add missing is_read column to threaded messages CTE

### LINKED ISSUE #27: Inbox never loads when "Group conversations" is enabled — backend crashes in a loop
After adding an email account and enabling the "Group conversations" option, the inbox stays empty and the backend keeps restarting. The sidebar shows the correct unread count but the message list never renders.

**Steps to reproduce**

1. Install and start MailFlow
2. Create an account and add an email account (Gmail via IMAP)
3. Enable "Group conversations" in the settings
4. Open the inbox

**Expected behavior**

Emails are displayed grouped by conversation thread.

**Actual behavior**

The inbox shows "Inbox is empty". The backend crashes immediately and restarts, repeating on every page load.

**Error from logs**

\`\`\`
severity: 'ERROR',
code: '42703',
routine: 'errorMissingColumn',
\`\`\`

PostgreSQL error `42703`: column `is_read` is undefined. Thrown when the backend executes `GET /api/mail/messages?threaded=true`.

**Version:** 1.0.2

saiththerobo:
Fixes #27.

The `is_read` column is used in the `COUNT(*) FILTER (WHERE NOT m.is_read)` window function inside the `ranked` CTE, but it was never included in the CTE's `SELECT` list. The outer query then references `is_read` from `ranked`, which PostgreSQL rejects with error `42703` (undefined_column).

This crashes the backend on every request to `GET /api/mail/messages?threaded=true` (triggered by the "Group conversations" setting), causing a restart loop and an empty inbox.

**Change:** add `m.is_read` to the inner CTE `SELECT` on line 174.

### DIFF
--- a/backend/src/routes/mail.js
+++ b/backend/src/routes/mail.js
@@ -171,7 +171,7 @@ router.get('/messages', async (req, res) => {
                COALESCE(m.thread_id, m.id::text) AS thread_id,
                m.subject, m.from_name, m.from_email,
                m.to_addresses, m.cc_addresses, m.reply_to, m.in_reply_to,
-               m.date, m.snippet, m.is_starred,
+               m.date, m.snippet, m.is_starred, m.is_read,
                m.has_attachments, m.account_id,
                a.name  AS account_name,
                a.email_address AS account_email,


saiththerobo:
Superseded by #34, which includes the same fix plus three additional bug fixes (thread expansion duplicates, folder context mismatch, migration bootstrap). Closing in favour of the combined PR.

PR #41, opened 2026-05-08
fix: thread display bugs (original sender/subject, badge count, folder context) and missing Cc field

### LINKED ISSUE #35: Thread list shows latest sender and subject instead of original
When viewing the threaded message list, the sender name, email, and subject displayed on each thread row belong to the **most recent reply** in the thread, not the original message that started it.

This is confusing because:
- The thread appears to be "from" the last person who replied, rather than whoever initiated the conversation
- The subject shown may be a modified reply subject (e.g. "Re: Re: Meeting tomorrow") rather than the original

**Expected:** The thread row should show the original sender and subject — the person who started the conversation and what it was about.

**Actual:** The thread row shows the sender and subject of the most recent message in the thread.

### LINKED ISSUE #36: Thread badge count does not match the number of messages shown on expansion
The message count badge on a thread row (e.g. "5") frequently does not match the number of messages actually shown when the thread is expanded.

This happens because the badge count and the expansion query use different scoping rules — the badge counts messages across all IMAP folders (INBOX, All Mail, Sent, etc.) while the expansion may show a different subset. For Gmail accounts in particular, the same message is stored under multiple folder paths, which causes double-counting or missing messages depending on which side has the mismatch.

**Expected:** The badge count should exactly match the number of message rows shown when the thread is expanded.

**Actual:** The counts diverge, most noticeably when viewing folders other than INBOX (e.g. All Mail), or on Gmail accounts where messages exist under multiple folder paths.

### LINKED ISSUE #37: CC recipients not shown in message header
When reading a message, the header only shows the **To** recipients. The **Cc** field is never displayed, even when the message has CC recipients.

This means there is no way to see who else received the message without opening the raw source.

**Expected:** If a message has CC recipients, they should appear in the message header below the To line.

**Actual:** Only the To line is shown. CC is always absent.

saiththerobo:
Fixes #35, #36, #37

Four independent bugs addressed in this PR, all affecting the threaded message view.

## Thread shows latest sender and subject instead of original (fixes #35)

The thread list row was displaying the `from_name`, `from_email`, and `subject` of the most recent message in the thread. Using `FIRST_VALUE(... ORDER BY date ASC)` window functions in the `ranked` CTE ensures the original sender and subject are always shown, matching the behaviour of Gmail and Outlook.

## Thread badge count does not match expansion (fixes #36)

Two separate problems caused the badge to diverge from the expansion:

- **Duplicate counting** — Gmail stores the same message under multiple IMAP folders (INBOX, All Mail, Sent, etc.). `COUNT(*)` over the raw rows inflated the count 2–6×. Fixed with a `deduped` CTE using `DISTINCT ON (account_id, thread_id, message_id)` before the window functions run.
- **Folder scope mismatch** — the badge counted messages across all folders while the expansion was filtered to a specific folder. Both are now scoped consistently: INBOX views filter to INBOX, all other folder views (All Mail, Sent, etc.) show the full thread across folders.

## Cc recipients not shown in message header (fixes #37)

The message header only rendered the To line. Cc recipients are now shown below To when present, on both desktop and mobile.

## Reading pane blank when clicking non-latest thread messages

When expanding a thread and clicking an older message, the reading pane showed "Select a message to read" because the message wasn't in the main list store. `MessagePane` now falls back to the `threadMessages` cache when the selected message isn't found in the main list.

### DIFF
--- a/backend/src/routes/mail.js
+++ b/backend/src/routes/mail.js
@@ -169,8 +169,12 @@ router.get('/messages', async (req, res) => {
     // thread_totals counts messages in the same folder scope as the list view so the
     // badge matches what the expansion will show. Unified inbox always scopes to INBOX;
     // per-account views use the requested folder (reuses $2 from filterValues).
+    // Only scope thread_totals to a specific folder for INBOX views, where the badge
+    // must match the expansion (which also filters to INBOX). For any other folder
+    // (All Mail, Sent, etc.) count across all folders so the badge reflects the true
+    // thread size rather than how many copies happen to be synced to that folder.
     const threadFolderFilter = (accountId && userAccountIds.includes(accountId))
-        ? `AND folder = $2`
+        ? (folder === 'INBOX' ? `AND folder = $2` : '')
         : `AND folder = 'INBOX'`;
     const threadResult = await query(`
       WITH deduped AS (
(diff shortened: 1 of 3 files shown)


maathimself:
Manually merged into main — changes applied directly to the branch and deployed. All three fixes (original thread sender/subject, badge count, CC display) are live.

PR #42, opened 2026-05-08
fix: improve message row highlight visibility in dark theme

saiththerobo:
The selected message row uses `--bg-elevated` as its background, which is nearly indistinguishable from the list background (`--bg-primary`) in the default dark theme. There is no clear visual indication of which message is currently open.

This PR replaces the selected row background with `--accent-glow` — a per-theme accent tint already defined for all themes in `themes.js` and used elsewhere in the UI — giving a subtle but clearly visible highlight on the active row that respects the user's chosen theme.

The same tint is applied as a "last viewed" indicator — when a message is opened and the user navigates away (e.g. switches folder), the row retains the tint until another message is selected, making it easy to find your place again.

Changes are in `MessageRow` and `ThreadRow` sub-rows in `MessageList.jsx`.

---

**Updated since review:**

- Reverted `--accent-glow` from structural surfaces (mobile header, search dropdown, bulk action bar, `EmptyState` icon containers) — those are back to `--bg-secondary` as they were
- Fixed `ThreadRow` on mobile: the last-viewed tint now shows correctly instead of being overridden by `--bg-primary`
- Added fix: tapping a multi-message thread on mobile now expands it for message selection instead of auto-opening the last message (discovered during testing, small addition to keep the mobile UX consistent)

### DIFF
--- a/frontend/src/components/MessageList.jsx
+++ b/frontend/src/components/MessageList.jsx
@@ -34,7 +34,7 @@ export default function MessageList() {
     selectedAccountId, selectedFolder, messages, setMessages,
     appendMessages, messagesTotal, setMessagesTotal, messagesOffset,
     setMessagesOffset, hasMoreMessages, setHasMoreMessages,
-    loadingMessages, setLoadingMessages, selectedMessageId,
+    loadingMessages, setLoadingMessages, selectedMessageId, lastViewedMessageId,
     setSelectedMessage, updateMessage, removeMessage,
     decrementUnread, incrementUnread, addNotification,
     searchQuery, setSearchQuery, isSearching, setIsSearching,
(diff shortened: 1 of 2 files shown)


saiththerobo:
this looks like the wrong solution

maathimself:
The `lastViewedMessageId` idea and the `--accent-glow` selection highlight are both great. The only issue is that `--accent-glow` got applied to a few structural surfaces it shouldn't touch — the mobile header, search dropdown, bulk action bar, and the `EmptyState` icon containers. Those should stay as they were.

There's also a small conflict on mobile — `ThreadRow` now forces `var(--bg-primary)` on mobile from a recent change, so the last-viewed tint won't show there. Worth a quick fix so it works consistently everywhere.

Get those sorted and this is good to merge!

PR #65, opened 2026-05-15
Feat: message list quick actions toggle

dcoffin88:
## Summary

Adds new settings UI for enabling or disabling hover-based action buttons.

The feature originally started as a hover delete button implementation, but once I started UI style matching I evolved to a broader hoverQuickActions setting to toggle all quick action buttons consistently for the message list UI.

## Changes

- Added settings UI for enabling/disabling hover quick actions
- Defaulted hoverQuickActions to disabled
- Updated action button rendering behavior to respect hover quick action settings
- Refined threading interaction behavior so action buttons remain accessible when threading is enabled
- Updated settings styling to match existing admin/message list configuration blocks

image

## Testing

- Verified hover quick actions correctly enable/disable from settings
- Confirmed default state is disabled
- Tested message list behavior with threading both enabled and disabled
- Verified action buttons render correctly in all combinations of hover quick actions on/off & threading on/off
- Checked styling consistency against existing settings components

---


maathimself:
Hey thanks for this, there are a few things to fix before I can merge it. Let me know.

**Bug — thread unread count**
When marking a thread as read from the hover buttons, the sidebar badge only drops by 1 instead of the full thread count. You will need to use the thread-aware handler instead of `handleMarkRead`.

**Default value**
Hover actions already existed for single message rows before this PR. Defaulting to off silently removes them for existing users — please default to `true` to preserve the current behaviour.

**Hardcoded strings**
The mark read and star tooltips in ThreadRow are hardcoded English. The translation keys already exist in the codebase (`contextMenu.markRead`, `contextMenu.markUnread`, `contextMenu.star`, `contextMenu.unstar`) — please use those.

**Translations**
A few of the On/Off labels came out as prepositions rather than state words (for example "Sur" in French, "En" in Spanish, "На" in Russian, "在" in Chinese). These will need a native speaker check or a second pass.

dcoffin88:
Thank you for your feedback.

**Bug - thread unread count**

- fixed with fix: thread-aware handleThreadMarkRead

**Default value**
- fixed from defaults to false unless localStorage already contains "true"
- now defaults to true unless localStorage already contains "false"

**Hardcoded strings**
- Sorry I missed these. This is fixed.

**Translations**
- flagged translations as translation required.

PR #71, opened 2026-05-17
Fix: threaded actions weren't catching all messages

dcoffin88:
## Summary

Fixes threaded row actions, so operations apply consistently across all messages in a thread where appropriate, while preserving the existing single-message behaviour.

## Changes

- Updated hover delete path to detect threaded rows.
- Kept single-message deletes on the original single-message API path.
- Updated thread rows to resolve all messages in the thread before applying mark read/unread.
- Updated star/unstar actions to resolve and apply to the full thread.
- Updated move-to-folder actions to resolve and apply to all messages in the thread.

## Testing

- Verified single-message deletes still use the existing single-message API path.
- Verified threaded rows resolve all messages before mark read/unread.
- Verified star/unstar applies across the full thread.
- Verified move-to-folder applies across all messages in the thread.
- Checked that non-threaded row behaviour remains unchanged.

---

### DIFF
--- a/frontend/src/components/ContextMenu.jsx
+++ b/frontend/src/components/ContextMenu.jsx
@@ -173,6 +173,8 @@ export default function ContextMenu({ x, y, message, onClose, onAction }) {
   const [customSnoozeView, setCustomSnoozeView] = useState(false);
   const [customDate, setCustomDate] = useState('');
   const [customTime, setCustomTime] = useState('09:00');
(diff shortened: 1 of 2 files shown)


maathimself:
Rebased onto main and merged manually — conflicts resolved by preserving the envelope icon changes from #68 alongside this PR's hasUnread logic.

PR #66, opened 2026-05-15
Feat: add customizable swipe actions when isMobile

dcoffin88:
## Summary

Adds customizable swipe actions for mobile message list interactions.

This feature introduces configurable swipe gestures when isMobile is enabled.

image

## Changes

- Added customizable swipe actions for mobile message lists
- Added translation support for swipe action labels/settings
- Implemented configurable swipe behavior for message actions
- Fixed swipe "mark as read" behavior/state handling

## Testing

- Tested all new swipe actions on mobile
- Verified swipe gestures trigger the correct actions
- Checked behavior with different swipe action configurations
- Checked that disable option works.

---

### Contributor License Agreement

By submitting this pull request I confirm that:

- [x] I have read and agree to the [Contributor License Agreement](../CLA.md).
- [x] My contribution is my own original work (or I have identified any
      third-party material and confirmed it is compatible with the CLA).
- [x] I have the right to submit this contribution under the terms of the CLA.

PR #74, opened 2026-05-18
feat: add reply and replyall to swipe options

dcoffin88:
## Summary

This restores code that was included in PR #66 following feedback from @saiththerobo, but was missed during rebase.

## Changes

- Add Reply action to swipe options
- Add Reply All action to swipe options

## Testing

- Verify swipe actions display correctly
- Verify Reply opens compose flow for sender only
- Verify Reply All includes all recipients correctly
- Confirm existing swipe actions continue to work as expected

---


maathimself:
One small thing before merge: --blue isn't defined in the theme system (unlike --green, --amber, etc. which all are), so the replyAll swipe background color won't adapt to different themes. Could you swap it for an existing variable or just hardcode the hex directly?

dcoffin88:
All set.

maathimself (approved):
lgtm

PR #75, opened 2026-05-18
Fix: message flicker on delete

dcoffin88:
## Summary

Fixes an issue where messages could briefly reappear in the message list after being deleted while the optimistic delete/undo flow was still pending.

This happened because background refreshes or sync events could reinsert messages before the delete operation fully completed.

## Changes

- Apply delete guards during message refreshes
- Preserve pending delete state across sync/list reloads
- Add completed delete handling for optimistic delete cleanup
- Ensure bulk delete flows correctly update delete guard state
- Fix pending delete cleanup during component unmount

## Testing

- Deleted single messages and verified they do not reappear during undo window
- Deleted threaded/bulk messages and verified consistent behavior
- Triggered sync/refresh events during pending deletes
- Verified undo restore behavior still works correctly
- Verified delete failure paths correctly clear guards and restore messages

---


maathimself (approved):
Lgtm

PR #78, opened 2026-05-18
Fix: message pane padding

dcoffin88:
## Summary

Emails with 100% width were rendering flush against the container border.

## Changes

- Added padding around email body content so email content has same visual consistency as other panes.

## Testing

- Verified spacing renders correctly in desktop and mobile email clients

---

### DIFF
--- a/frontend/src/components/MessagePane.jsx
+++ b/frontend/src/components/MessagePane.jsx
@@ -1174,6 +1174,7 @@ export default function MessagePane() {
           )}
           <div style={{
             position: 'relative',
+            padding: '14px 16px 12px',
             background: 'white',
             borderRadius: isMobile ? 0 : 10,
             border: isMobile ? 'none' : '1px solid var(--border-subtle)',


maathimself:
One thing, on mobile the padding is applied unconditionally, but the existing mobile styles in that block all strip the card chrome on small screens (no border, no border radius, and the outer wrapper already sets zero horizontal padding on mobile). Adding 16px each side on mobile goes against that pattern and will squeeze email content on small screens. A quick ```isMobile ? 0 : '14px 16px 12px'``` should sort it. Let me know what you think.

dcoffin88:
Correct, the existing isMobile padding is untouched and works great, but in PR #78 and PR #79 I'm fixing the DIV inside the container to align the actual content with the padding that exists in the header and elsewhere.

**Before:**
_Desktop_
image
_Mobile_
image

**After:**
_Desktop_
image
_Mobile_
image

### And for PR #79

**Before:**
_Desktop_
image
_Mobile_
image

**After:**
_Desktop_
image
_Mobile_
image

maathimself (approved):
lgtm

maathimself:
You are correct, thanks.

Issue #81, opened 2026-05-18
Sender avatars: per-sender color instead of account color


saiththerobo:
## Feature idea

Right now the sender avatar in the message list shows the first letter of the sender's name with the color of the account it was received on. This works fine but it makes all messages look visually identical — every avatar in an inbox is the same color.

Gmail and a few other clients assign a consistent color per sender (derived from the sender's address or name), so you get a bit of visual variety in the list and can start to recognise senders at a glance without reading the name.

## What this could look like

- Hash the sender's email address to pick a color from a fixed palette
- The first letter stays the same, only the background color changes per sender
- Same sender always gets the same color (deterministic), regardless of which account the mail came in on

## Questions / things to discuss

- Is this something worth pursuing, or is the current behavior intentional?
- Should the account color still take priority in some views (e.g. unified inbox)?
- Any preference on palette size / approach?

Happy to hear thoughts — not sure if this fits as an issue or would be better as a discussion.

maathimself:
Yes, I like it. I was going to suggest making it a setting that the user can configure to their preference, but I think the cleaner path is to just go with per-sender colors as the new default (since it's objectively more informative when the account dot already handles account identification). Is this something you want to tackle?

saiththerobo:
oh yes can give it a shot 😄

PR #87, opened 2026-05-19
feat: per-sender avatar colors

saiththerobo:
## Summary

Implements per-sender avatar colors as discussed in #81. Each sender gets a consistent color derived from their email address, so the message list shows visual variety and senders become recognisable at a glance without reading the name. The small account-color dot used for account identification in the unified inbox is unchanged.

## Changes

- Added a 16-color palette to `themes.js` covering the full hue wheel with sufficient contrast for white text
- Added `senderColor(email)` in `themes.js`: hashes the email address to a palette index — deterministic, no storage, consistent across devices and browsers
- Updated the two avatar sites in `MessageList.jsx` (compact threaded view and wide non-threaded view) and the one in `MessagePane.jsx` to use `senderColor` instead of `account_color`

## Testing

- Tested against a real inbox with multiple accounts
- Verified that the same sender always receives the same color across page reloads and devices
- Verified the account-color dot is unaffected

---

### DIFF
--- a/frontend/src/components/MessageList.jsx
+++ b/frontend/src/components/MessageList.jsx
@@ -4,6 +4,7 @@ import { useStore } from '../store/index.js';
 import { api } from '../utils/api.js';
 import { format, isToday, isYesterday, isThisYear } from 'date-fns';
 import { LAYOUTS } from '../layouts.js';
+import { senderColor } from '../themes.js';
 import { useMobile } from '../hooks/useMobile.js';
 import ContextMenu from './ContextMenu.jsx';
 import { shortcutBus } from '../utils/shortcutBus.js';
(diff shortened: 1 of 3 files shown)


maathimself:
### Thank you, the palette looks good and the collision-free assignment logic is clean.

A couple things before this can merge:

**Performance:** `senderColor()` currently parses localStorage on every call, and it's called once per visible row. On a 50-message page that's 50 synchronous JSON parses per render. Please add a module-level Map as an in-memory cache, read from localStorage once on first call, write back only when a new sender is assigned, and serve all subsequent lookups from the Map.

**Cross-device consistency**: It says the same sender always gets the same color, that's only true within a single browser. On a different device or browser the `unused[0]` fallback will land on a different slot depending on what order senders were first seen. If that's acceptable, fine, but it's worth being explicit about. If you want true determinism across devices, a pure hash would get you there without localStorage at all.

Let me know.

saiththerobo:
Went with pure hash in the end — dropped the localStorage approach entirely. `senderColor` is now just two lines: hash the email, index into the palette. No storage, no cache, works identically on every device and browser.

Looking back the collision-avoidance logic was probably overthinking it — 16 colors across a typical inbox is enough that clashes are rare, and the cross-device inconsistency was a worse property to silently have than an occasional shared color.

maathimself (approved):
LGTM

PR #108, opened 2026-05-22
fix: restore text selection visibility in Parchment theme

### LINKED ISSUE #105: Minor text highlight issue in "Parchment" theme
Its not a priority but since I love Parchment theme I just wanted to report an issue with text selection highlighter in compose is blended in with background, making it completely look like no text is selected. the text in the screenshot is actually selected but looks like nothing has been selected at all.

Image

saiththerobo:
Closes #105

## Summary

The Parchment theme's cream background (`#f5f0e8`) causes the browser's default text selection highlight to blend in and become nearly invisible. This fix injects a theme-specific `::selection` colour when the Parchment theme is active.

## Changes

- Added `--selection-bg: rgba(139,69,19,0.25)` to the Parchment theme vars — a translucent saddle brown that matches the theme's accent colour
- Extended `applyTheme()` to inject a `<style id="theme-selection-style">` tag into `<head>` when a theme defines `--selection-bg`, and clear it when switching away

A pure CSS approach (`::selection { background-color: var(--custom-prop) }`) was attempted first but CSS custom properties do not reliably inherit into `::selection` across browsers. Dynamic style injection is the standard workaround for this.

Other themes are unaffected — they continue using the browser's native selection colour.

## Testing

Tested in Parchment theme: selected text in compose, message body, and search input is now clearly highlighted with a warm brown tint. Switching to any other theme restores the default browser selection colour.

Screenshot From 2026-05-22 10-38-06

---


maathimself:
Hey, I think the approach here is more complex than it needs to be. The `applyTheme` loop already sets every entry in theme.vars as a CSS custom property on `:root`, so `--selection-bg` ends up available in CSS automatically. That means we don't need the `<style>` injection at all — we can just update `index.css` line 91 to use it as a fallback:

`::selection { background: var(--selection-bg, var(--accent-dim)); color: var(--text-primary); }`

That's two lines total (one in `themes.js` to add the var, one in `index.css`), no DOM mutation, and it follows the same pattern every other theme property uses. The `data-theme` attribute also doesn't appear to be used anywhere — happy to drop that too.

Would you mind updating the PR with that approach or let me know what you think? Otherwise happy to close this and apply the fix directly. Let me know

saiththerobo:
Would you be able to pick this up, I tried the CSS var approach but ran into issues getting it to work reliably

PR #110, opened 2026-05-22
feat: font family and size picker in composer

### LINKED ISSUE #104: Text size selection option in composer
Please add a font text size selector in email composer. please make font style and size selection persistent if possible.

saiththerobo:
Closes #104

## Summary

Adds font family and font size selectors to the compose toolbar. Both selections are persisted in localStorage so the composer remembers your preferences across sessions.

## Changes

- New `FontSize` TipTap extension (extends `TextStyle`) with `setFontSize` / `unsetFontSize` commands
- Font family picker: Sans Serif (system default), Arial, Georgia, Times New Roman, Courier New, Verdana, Trebuchet MS — defaults to Sans Serif
- Font size picker: 10–36px in common increments — defaults to 14px (standard email body size)
- Both pickers reflect the active format when text is selected
- Both persist last-used values to localStorage and restore them on next compose open
- On new compose (not reply/forward), saved preferences are pre-applied so typed text immediately uses the chosen font and size

## Testing

- Open compose, change font and size, send — verify the email renders with the chosen formatting
- Close and reopen compose — verify both dropdowns restore the saved values
- Select text with mixed formatting — verify dropdowns reflect the active format at the cursor
- Reply/forward — verify saved preferences are shown in the dropdowns but not forced onto the quoted content

---


maathimself:
Hey, thanks for this, font size persistence in the composer is a great addition and the TipTap extension itself is correctly implemented. A few things I think need to be addressed before we merge:

**Must fix**

1. **`localStorage` key names** the project uses a `mailflow_` prefix for all keys (`mailflow_page_size`, `mailflow_scroll_mode`, etc.). Please rename `compose-font-size` → `mailflow_compose_font_size` and `compose-font-family` → `mailflow_compose_font_family`.

2. **`onChange` vs `onMouseDown` on the size `<select>`** every other toolbar control uses `onMouseDown` + `e.preventDefault()` to keep the editor focused before applying the mark. Using `onChange` can cause the editor to lose its selection before the font size is applied, so it may end up formatting nothing. Please match the existing pattern.

3. **Missing `useEffect` deps** `isReply` and `isForward` are used inside the effect that pre-applies saved preferences but are omitted from the dependency array. Please add them.

**Minor**

- The font family label rename from `Default` → `Sans Serif` is a user-visible change and `Sans Serif` is slightly misleading (the actual rendered font depends on the OS/browser default). Worth a deliberate decision either way.
- When the cursor moves to unstyled text after a preference change, the toolbar select shows the mount-time saved value rather than the current one produces a subtle mismatch. Low priority but worth a note.

Let me know.

maathimself:
The font family `<select>` still uses plain `onChange` with no selection snapshot. You correctly fixed font size with `savedSelectionRef` + `onMouseDown`, but the family picker wasn't updated to match. Applying a font family to a text selection will still silently operate on a collapsed cursor. Please apply the same `onMouseDown` + `savedSelectionRef` pattern to the family picker and you're good to merge.

maathimself:
I added the final fix, then merged. Thank you

PR #111, opened 2026-05-22
feat: allow local and self-signed SSO providers

### LINKED ISSUE #109: Can't set up local PocketID as SSO method
**Describe the bug**
I'm trying to set up my (LAN-only) PocketID as an SSO provider, but app doesn't allow that for some reason.

**To Reproduce**
1. Go to Settings -> SSO
2. Click on "Add"
3. Select PocketID as provider
4. Set issuer url to "https://pocketid.internal/" (or any other non-global SNI)
5. See error "Issuer URL: Host cannot be a local address".

**Expected behavior**
SSO provider created

**Screenshots**
Image

**Desktop (please complete the following information):**
 - OS: Linux (but no matter)
 - Browser Firefox
 - Version 140 ESR

saiththerobo:
Closes #109

## Summary

Adds an **Allow local / self-signed connections** toggle to the SSO provider form. When enabled, the provider bypasses the SSRF host validation (which blocks `.local`, `.internal`, private IPs, etc.) and skips TLS certificate verification. This allows self-hosted LAN-only OIDC providers like PocketID to be used without needing a globally-routable hostname or a CA-signed certificate.

The toggle defaults to off — existing providers and the validation behaviour for public providers are unchanged.

## Changes

- `0010_oidc_allow_insecure.sql` — adds `allow_insecure` column to `oidc_providers` (default false)
- `backend/src/routes/admin.js` — GET/POST/PATCH routes accept and persist `allow_insecure`; host validation and HTTPS enforcement are skipped when it is set
- `backend/src/routes/oidc.js` — `getDiscovery()` skips host/HTTPS checks and uses an insecure `https.request`-based fetch (bypasses TLS cert verification) when `allow_insecure` is set; same applies to the token exchange and JWKS fetch (via `jose`'s `customFetch` option)
- `frontend/src/components/AdminPanel.jsx` — amber toggle with a warning description, wired into the existing provider form
- `frontend/src/locales/*.json` — `allowInsecure` / `allowInsecureDesc` strings added to all 7 locales

## Testing

Tested end-to-end with a local Dex instance running over plain HTTP on a private LAN IP (`http://192.168.x.x:5556`) — a host that would previously have been rejected by both the private-IP check and the HTTPS requirement. With the toggle on, the provider saved, discovery completed, and SSO login succeeded.

---


saiththerobo:
Tested end-to-end using a local [Dex](https://github.com/dexidp/dex) instance (plain HTTP over a private LAN IP), which hits both the private-IP block and the HTTPS requirement that were previously rejecting PocketID. With the toggle on, discovery, token exchange, and login all completed successfully. I don't have a PocketID instance to test against directly — would appreciate confirmation from the reporter that this resolves the original issue.

maathimself:
Hey, overall this looks good. A few things to address before we merge:

**Must fix**

1. **Discovery cache keying** `discoveryCache` is keyed on `issuerUrl` alone. If the `allow_insecure` flag is toggled, the stale JWKS object (built with the wrong fetcher) will be served for up to 5 minutes. Please key the cache on `${issuerUrl}:${allowInsecure}` or clear the entry on PATCH.

2. **Token exchange timeout dropped unnecessarily** `signal: provider.allow_insecure ? undefined : AbortSignal.timeout(10000)` removes the 10-second timeout on the token endpoint request for insecure providers. Being insecure and being slow are orthogonal please keep the timeout unconditionally.

3. **Jose's abort signal ignored for JWKS fetches** `makeInsecureFetch(signal)` correctly closes over the discovery `AbortSignal`, but when jose calls the returned function for JWKS it passes its own `signal` in the options object, which is silently dropped. The JWKS fetch ends up running on whatever time is left from the discovery signal. Please handle `opts?.signal` inside the returned function.

**Low priority**

- **PATCH partial-update validation** the HTTPS/host check reads `allow_insecure` from the request body. A partial PATCH that omits `allow_insecure` (e.g. just updating the name) will see it as `undefined` (falsy) and incorrectly reject a provider that already has `allow_insecure = true` stored. Consider reading the stored value from the DB before validating, or requiring the client to always send it.

- **Trailing newlines** the 6 non-English locale files lost their trailing newlines, which will create noise in future diffs.

Security posture looks good the flag is correctly admin-only, `rejectUnauthorized: false` is scoped tightly, issuer mismatch checking is preserved, and the UI warning copy is clear. Once the three must-fix items above are addressed this is ready to go. Thanks

maathimself (approved):
Looks all good

PR #129, opened 2026-06-01
feat: print email with customizable Ctrl+P shortcut

saiththerobo:
## Summary

Adds a print action that opens a formatted print view of the current message, wired into the existing customizable shortcut system as a first-class action (default Ctrl+P / Cmd+P on Mac).

## Changes

**Print feature**
- `MessagePane`: `handlePrint` builds a self-contained HTML page with the message header and body, opens it in a new window, and calls `window.print()`. HTML is XSS-escaped before injection.
- Toolbar print button added alongside Reply / Forward / Archive.
- i18n strings added for all supported locales.

**Shortcut system integration**
- `defaultShortcuts.js`: added `printMessage` action (`defaultKey: 'ctrl+p'`), `parseModKey` helper, `buildModKeyMap` for modifier+key dispatch, and `modLabel` / `modCompactLabel` for platform-aware display (⌘ on Mac, Ctrl / ^ elsewhere).
- `MailApp.jsx`: global key handler now checks `modKeyMap` for Ctrl/Cmd combos — only intercepts keys explicitly registered, passes everything else through to the browser unchanged.
- `MessagePane.jsx`: raw `document.addEventListener` for Ctrl+P replaced by `shortcutBus.on('printMessage', ...)`, consistent with all other pane shortcuts.
- `AdminPanel.jsx`: shortcuts settings tab renders and records modifier combos (e.g. `Ctrl + P`); users can rebind print to any key or modifier combo.
- Toolbar badges (Reply, Forward, Archive, Print) are now dynamic — they update live when the user rebinds a shortcut, and disappear if the shortcut is cleared.

## Testing

- Print opens a correctly formatted view (subject, from, to, cc, date, body) and triggers the browser print dialog.
- Ctrl+P / Cmd+P intercepted; browser print dialog suppressed.
- Shortcut appears in the `?` help overlay and the Settings → Shortcuts tab.
- Rebinding print to another key (e.g. Ctrl+I) updates the toolbar badge and dispatches correctly.
- Clearing the print shortcut hides the toolbar badge.
- All pre-existing shortcuts (r, f, e, j, k, gi, etc.) continue to work.
- Unregistered Ctrl/Cmd combos (Ctrl+C, Ctrl+V, Ctrl+Z, Cmd+Tab, etc.) pass through to the browser unaffected.

---


saiththerobo:
**Additional fix: `R` shortcut now always triggers plain reply**

While testing, I noticed an inconsistency in how the `reply` action was wired. The `paneActionsRef` had:

```js
reply: () => handleReply(defaultReplyAll),
```

This meant that if the user set their default reply mode to "Reply All" in settings, pressing `R` would open a Reply All compose window — even though `R` is defined and labelled as the *Reply* shortcut. Both `R` and `A` would silently do the same thing, which is confusing.

The fix changes `reply` to always call `handleReply(false)`:

- `R` → always plain reply
- `A` → always reply all
- Toolbar button → follows the default setting (unchanged)

The default reply preference now only controls which action the toolbar button performs, not what the keyboard shortcuts do.

maathimself:
Hey, one thing to fix before merging:

**Regression in `MessagePane.jsx`:**

```js
// This PR changes:
reply: () => handleReply(defaultReplyAll),
// back to:
reply: () => handleReply(false),

saiththerobo:
done

maathimself (approved):
Looks good

PR #144, opened 2026-06-03
feat: admin-controlled policy for self-hosted mail server connections

### LINKED ISSUE #126: IMAP: Host cannot be a private or reserved IP address
I love to use my self hosted protonmail-bridge and don't want to use external addresses. 

Is it possible?

saiththerobo:
## Summary

Closes #126. Also addresses the use case raised in discussion #116.

Adds three admin-only toggles under Settings → Security → **Mail Server Connection Policy**, allowing self-hosted deployments (protonmail-bridge, Dovecot on localhost, etc.) to connect to local IMAP/SMTP servers while keeping the existing SSRF protection intact for internet-facing instances.

All three flags default to off, so existing deployments are unaffected — **with one exception noted below**.

## ⚠️ Breaking Change

**Accounts with "Skip TLS verification" enabled will lose connectivity after upgrading** until an admin enables **Allow insecure TLS** in Settings → Security → Mail Server Connection Policy.

This affects any account configured to connect to a local IMAP server with a self-signed certificate. Enable the toggle before or immediately after deploying to restore connectivity.

## Changes

- `backend/migrations/0014_account_allow_private_host.sql` — adds `allow_private_hosts`, `allow_insecure_tls`, `allow_nonstandard_ports` to `system_settings`
- `backend/src/services/connectionPolicy.js` — new module that reads the three flags from the DB, with a 30s TTL cache to avoid a DB hit on every IMAP connection; exposes `invalidateConnectionPolicyCache()` called from PATCH `/settings` so changes take effect immediately
- `backend/src/services/hostValidation.js` — all exports accept `{ allowPrivate }` option; DNS rebinding protection is preserved
- `backend/src/services/imapManager.js` — reads policy on every connection; enforces TLS requirement; policy errors surface to the UI via `sync_error` and WebSocket broadcast
- `backend/src/routes/accounts.js` — several fixes alongside the policy enforcement:
  - PATCH `/accounts/:id` now accepts `imap_host`, `imap_port`, `smtp_tls` (previously locked after creation despite being shown in the edit form)
  - `imap_tls` is auto-derived from port on create and edit (`port % 1000 === 993` → TLS, otherwise plain) — no user-visible field
  - `imap_skip_tls_verify` is now correctly saved at account creation (was missing from the INSERT) and on edit (was missing from the PATCH allowlist)
  - `imap_tls`, `smtp_tls`, `imap_skip_tls_verify` added to `SAFE_FIELDS` and the GET query so the edit form loads correct values
  - IMAP host/port changes trigger automatic reconnect
  - invite-email SMTP fallback path now reads system policy (fixes dead `account.allow_private_host` field reference and missing TLS policy enforcement)
- `backend/src/routes/admin.js` — PATCH `/settings` persists the three flags and invalidates the policy cache
- `backend/src/routes/send.js` — enforces policy for outgoing SMTP
- `frontend/src/components/AdminPanel.jsx` — policy toggles in Security tab (auto-save on click); account form shows skip-TLS toggle only when `allowInsecureTls` is enabled; `handleEdit` now sends IMAP/SMTP fields that were previously silently dropped; `imap_skip_tls_verify` initialised to `false` in new account form
- `README.md` — Upgrading section documents the breaking change
- Locale files updated for en, de, es, fr, it, ru, zhCN

## Testing

- All existing tests pass (224 backend + 791 i18n)
- New unit tests for `makeClientCfg` TLS enforcement and `rejectUnauthorized` under all policy/account combinations
- New `allowPrivate` tests for all three `hostValidation` exports
- Manually tested end-to-end with Greenmail (local IMAP/SMTP server):
  - Plain IMAP (port 3143) connects when policy allows, fails with UI-visible error when revoked
  - IMAPS with self-signed cert (port 3993) connects with skip-TLS-verify on, fails when policy is off
  - `imap_skip_tls_verify` correctly persists at account creation and via edit
  - IMAP host/port editable after creation and reconnects automatically

---


maathimself:
Hey @saiththerobo — two things to fix before merging:

1. **`admin.js` dead code** — `account.allow_private_host` doesn't exist in the schema, so `allowPrivate` is always `false` in the test-email-send path. Replace with the system policy like the rest of the PR.

2. **Breaking change for `imap_skip_tls_verify = true` accounts** — these accounts will lose connectivity on deployment until an admin enables "Allow insecure TLS". Needs a callout in the migration guide or release notes.

Also consider caching `getConnectionPolicy()` — it hits the DB on every IMAP connection.

saiththerobo:
Note on the failing CI: the frontend audit step is failing on `esbuild ≤0.28.0` (high severity, GHSA-67mh-4wv8-2f99). This is pre-existing — it also fails on upstream `main` (run #27442680541). The fix requires bumping Vite from v5 to v8 which is a breaking change; happy to tackle it in a follow-up if you'd like to keep it out of scope for this PR.

maathimself:
The Upgrading section should also note that `ALLOW_PRIVATE_IMAP_HOSTS=true` no longer bypasses host validation — deployments using it to connect to local mail bridges will fail after upgrading and will need to enable **Allow private / local hosts** in Settings → Security instead.

maathimself (approved):
looks good

PR #145, opened 2026-06-03
i18n: fill translation gaps + add coverage test

saiththerobo:
## What's in here

I noticed a handful of missing keys across the non-English locales (some strings silently falling back to English at runtime) and a few hardcoded user-visible strings that weren't going through `t()` at all. Went down a rabbit hole fixing them and ended up writing a test so it doesn't drift again.

**New file: `frontend/src/locales/i18n.test.js`**
Zero-dependency test (Node built-in `node:test`). Four suites, run with:
```
cd frontend && node --test src/locales/i18n.test.js
```
1. **Source coverage** — every locale key must be referenced in source (catches dead keys before they spread across 7 locale files)
2. **Key coverage** — every key in `en.json` must exist in every other locale
3. **Value uniqueness** — no two locales may share the same translation for a key (catches untranslated strings), with an explicit whitelist for brand names / legitimate cross-language overlap
4. **Hardcoded strings** — scans JSX for user-visible text not wrapped in `t()`

**Locale fixes:**
- ~50 missing keys filled across de/es/fr/it/ru/zhCN
- 11 dead keys removed from all locale files
- 18 new keys added (compose toolbar tooltips, empty-state strings, aria-labels, etc.)

**Component fixes:**
- `ComposeModal`, `MessageList`, `AdminPanel`, `MessagePane`, `NotificationToasts` — hardcoded strings replaced with `t()` calls

No behaviour changes, no new dependencies. Happy to split it if you'd prefer locale fixes and the test as separate PRs — just let me know.

---
*758 tests, 0 failures on current `main`.*


saiththerobo:
Closing — this had changes from other open PRs mixed in. Clean version coming up.

PR #146, opened 2026-06-03
i18n: fill translation gaps + add coverage test

saiththerobo:
## Summary

Found missing translations across the non-English locales (keys silently falling back to English at runtime) and hardcoded user-visible strings not going through `t()`. Fixed both, and added a test so it doesn't drift again.

## Changes

**New: `frontend/src/locales/i18n.test.js`**
Zero-dependency test (Node built-in `node:test`). Four suites:
1. **Source coverage** — every locale key must be referenced in source (catches dead keys before they spread across 7 locale files)
2. **Key coverage** — every key in `en.json` must exist in every other locale
3. **Value uniqueness** — no two locales may share the same translation for a key (catches untranslated strings), with a whitelist for brand names and legitimate cross-language overlap
4. **Hardcoded strings** — scans JSX for user-visible text not wrapped in `t()`

**CI (`frontend/package.json`, `.github/workflows/ci.yml`)**
Added `"test": "node --test"` script and a Test step to the frontend CI job, matching the backend pattern. The test runner auto-discovers `*.test.js` files, so future test files are picked up automatically.

**Locale fixes:**
- ~50 missing keys filled across de/es/fr/it/ru/zhCN
- 11 dead keys removed from all locale files
- 18 new keys added (compose toolbar tooltips, empty-state strings, aria-labels, etc.)

**Component fixes:**
- `ComposeModal`, `MessageList`, `AdminPanel`, `MessagePane`, `NotificationToasts` — hardcoded strings replaced with `t()` calls

## Testing

747 tests, 0 failures — verified locally via `npm test` and in CI.

---


maathimself (approved):
looks good

PR #147, opened 2026-06-04
chore: add ESLint to frontend and backend

saiththerobo:
## Summary

Adds ESLint to both the frontend and backend with a `no-undef` error rule that catches undefined variable references before they reach production. This is the class of bug that caused a `ReferenceError: t is not defined` crash in the compose modal when `t` from `useTranslation()` was used in `RichToolbar` without being declared there.

Also fixes a pre-existing `no-undef` the linter found in the backend: `adjustFolderCounts` was defined in `routes/mail.js` but called in `services/imapManager.js` without being imported, causing a silent `ReferenceError` on snooze wakeup. Moved to `utils/mailUtils.js` and imported in both files.

## Changes

- `frontend/eslint.config.js` — ESLint config with `no-undef: error`, quality rules as warnings (`no-unused-vars`, `no-empty`, `no-useless-assignment`, `react-hooks/rules-of-hooks`, `react-hooks/exhaustive-deps`)
- `backend/eslint.config.js` — same approach for Node, using `globals.node` instead of `globals.browser`
- `frontend/package.json`, `backend/package.json` — `lint` script added; ESLint packages in `devDependencies` only (not installed in production image)
- `.github/workflows/ci.yml` — `Lint` step added to both backend and frontend jobs, runs after `npm ci`
- `backend/src/utils/mailUtils.js` — `adjustFolderCounts` moved here and exported
- `backend/src/routes/mail.js` — imports `adjustFolderCounts` from `mailUtils.js` instead of defining it locally
- `backend/src/services/imapManager.js` — imports `adjustFolderCounts` from `mailUtils.js`

## Testing

- Confirmed `npm run lint` exits 0 on the current codebase (0 errors, warnings only)
- Confirmed the linter catches `no-undef` on a version of `ComposeModal.jsx` without the `useTranslation` fix (10 errors reported)
- Confirmed the `adjustFolderCounts` calls in `imapManager.js` previously produced 2 `no-undef` errors; fixed and verified clean

---

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,6 +24,10 @@ jobs:
         working-directory: backend
         run: npm ci
 
+      - name: Lint
+        working-directory: backend
+        run: npm run lint
+
       - name: Syntax check
         working-directory: backend
         run: node --check src/index.js
@@ -52,6 +56,10 @@ jobs:
       - name: Install dependencies
         working-directory: frontend
         run: npm ci
+
+      - name: Lint
+        working-directory: frontend
+        run: npm run lint
 
       - name: Build
         working-directory: frontend
--- a/backend/eslint.config.js
+++ b/backend/eslint.config.js
@@ -1,0 +1,27 @@
+import js from '@eslint/js'
+import globals from 'globals'
+
+export default [
+  js.configs.recommended,
+  {
+    files: ['**/*.js'],
+    rules: {
+      // Crash-causers — block CI
+      'no-undef': 'error',
+      // Quality issues — report but don't block
+      'no-unused-vars': 'warn',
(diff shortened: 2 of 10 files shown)


saiththerobo:
Note on the warnings: the rules set to `warn` (unused vars, empty catch blocks, useless assignments, react-hooks exhaustive-deps) are not noise — I consider them real issues worth fixing. They're warnings rather than errors to keep CI green on the existing codebase while this is introduced, rather than making this PR also touch 80+ unrelated lines across multiple files. The intent is to tighten them to `error` in a follow-up once the existing violations are cleaned up.

maathimself:
Applied directly to main with the invalid 'preserve-caught-error' rule removed. Thanks @saiththerobo!

PR #160, opened 2026-06-15
fix: prevent virtual folders from stealing messages during sync

### LINKED ISSUE #159: Navigating to All Mail causes inbox to appear empty
## Describe the bug

After navigating to the `[Gmail]/All Mail` folder for the first time, going back to the inbox shows it completely empty. The messages come back after manually triggering a sync.

This seems to happen because visiting All Mail triggers a background sync for that folder, and something in that sync process is displacing the inbox messages in the database. The inbox then shows 0 messages until the next INBOX sync re-creates them.

## Steps to reproduce

1. Add a Gmail account and let the inbox sync normally
2. Navigate to `[Gmail]/All Mail`
3. Navigate back to inbox
4. Inbox appears empty
5. Manually sync — messages reappear

## Expected behavior

Inbox remains unaffected when syncing other folders.

## Environment

- Gmail account via IMAP
- Conversation view enabled

saiththerobo:
## Summary

Fixes #159.

When syncing `[Gmail]/All Mail` for the first time, the message relocation logic displaces INBOX messages — updating their `folder` to `[Gmail]/All Mail` in the DB. The inbox then appears empty until the next INBOX sync recreates the rows.

The relocation step handles messages genuinely moved between folders (e.g. INBOX → Trash). It uses a `COUNT=1` guard to avoid merging two existing copies of the same message. On the first sync of a Gmail virtual folder though, only the INBOX copy exists (`COUNT=1`), so the guard passes and the INBOX row gets stolen.

The fix skips relocation when the destination folder has `special_use = '\All'` or `'\Important'`, since these are virtual folders where a message can coexist with copies in real folders.

## Changes

- `backend/src/services/imapManager.js`: add `special_use` check to the relocation query so virtual folders cannot displace messages from real folders

## Testing

1. Add a Gmail account and let the inbox sync
2. Navigate to All Mail for the first time
3. Navigate back to inbox — messages remain intact (previously they disappeared until manual sync)

---


maathimself:
There's a second identical relocation query in the backfill path at line 1558 that has the same vulnerability. Can you apply the same `special_use` check there too?

maathimself (approved):
lgtm

PR #161, opened 2026-06-15
fix: show sent replies in conversation threads and live-update on send

### LINKED ISSUE #157: missing replies in conversation view (inbox only)
cant figure out when setting to conversation view, to have it include replies/sent mail from the same email thread when browsing the inbox

**Describe the solution you'd like**
include replies in the collapsible conversation view for that particular message. it appears to only show INCOMING emails in the convo stack when viewing from inbox folder
but when i go into the sent folder, and expand the stack, all of the sent&received messages are there

perhaps this is intended functionality, or i have missed a setting somewhere?

thanks!

Image

Image

saiththerobo:
## Summary

Fixes #157.

This PR addresses two related gaps in the conversation view:

1. **Sent replies missing from the conversation** — thread expansion was filtering messages to the current folder, so a reply in the Sent folder never appeared alongside the inbox messages it was part of.

2. **Reply doesn't appear in the open thread until manual refresh** — after sending a reply, the count badge incremented but the new message wasn't added to the open expansion, and an open message body would close as the list refreshed.

### How the fix works

- Thread expansion now shows every message in the conversation except those in trash or archive folders, regardless of which folder the user is viewing from — same logic for the badge count.
- After a reply is sent, the frontend tracks the affected thread key and on each `sync_complete` re-fetches that thread, appending only newly seen messages without disturbing existing references. The currently expanded thread is also refreshed in case incoming mail arrived in it.
- The user's own sent mail now goes through body prefetch, so its snippet shows up in the conversation preview without waiting for a manual click.

### Backend

- `routes/mail.js` — thread expansion excludes trash/archive instead of filtering to the current folder
- `services/messageService.js` — `thread_totals` and `deduped` CTEs apply the same trash/archive exclusion
- `services/imapManager.js` — `prefetchNewMessageBodies` now also runs on already-read newly inserted messages (e.g. user's own sent mail), and broadcasts `sync_complete` once a fresh snippet is generated so previews can refresh
- `services/messageService.test.js` — updated/extended threaded mode tests

### Frontend

- `components/ComposeModal.jsx` — dispatches `mailflow:reply-sent` with the thread key after a reply is sent
- `components/MessagePane.jsx`, `components/MessageList.jsx` — pass `message.thread_id` as `threadKey` when opening compose for a reply
- `components/MessageList.jsx` — tracks the just-replied thread via a ref, extracts a `mergeServerThread` helper that on every `mailflow:refresh` re-fetches the replied and currently expanded threads and merges in new messages (and previously empty snippets) without disturbing existing message references; preserves the open message in `threadMessages` when the refresh would otherwise drop it from the list

## Testing

1. Add an email account and exchange a few messages in a thread
2. Enable conversation view
3. Open the thread from inbox — sent replies appear alongside received messages and the thread count badge reflects the full thread size
4. With the thread open, send a reply — the new message appears in the sub-list automatically, the open message body stays put, and the thread auto-expands if it wasn't already
5. While the thread is expanded, simulate an incoming reply (e.g. from another device) — the new message appears without a manual refresh

---

### DIFF
--- a/backend/src/routes/mail.js
+++ b/backend/src/routes/mail.js
@@ -142,12 +142,16 @@ router.get('/thread/:threadId', async (req, res) => {
     const userAccountIds = accountsResult.rows.map(r => r.id);
     if (!userAccountIds.length) return res.json({ messages: [] });
 
-    // Only restrict expansion to INBOX when viewing the INBOX — ensures the expansion
-    // matches what the list shows. For All Mail and any other folder show all messages
-    // regardless of which folder they were synced under (All Mail backfill is skipped so
-    // not every message has a [Gmail]/All Mail row in the DB).
-    const folderFilter = (folder === 'INBOX') ? `AND m.folder = $3` : '';
-    const params = (folder === 'INBOX') ? [userAccountIds, threadId, folder] : [userAccountIds, threadId];
+    const accountsWithMappings = await query(
+      'SELECT id, folder_mappings FROM email_accounts WHERE id = ANY($1)',
+      [userAccountIds]
+    );
+    const excludedPaths = (await Promise.all(
+      accountsWithMappings.rows.flatMap(a => [
+        resolveTrashFolder(a.id, a.folder_mappings),
+        resolveArchiveFolder(a.id, a.folder_mappings),
+      ])
+    )).filter(Boolean);
 
     const result = await query(`
       WITH deduped AS (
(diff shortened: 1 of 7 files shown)


maathimself:
Three issues to fix before merging:

**1. `pendingRepliedThreadRef` never cleared** — after the first `mergeServerThread` call the new message is in `existingMap`, so `newOnly.length` is always 0 and the ref leaks permanently. Every subsequent sync fires an extra `getThread` call indefinitely. Fix: clear when `snippetFilled || newOnly.length > 0`.

**2. Prefetch threshold regression** — changing the guard from `newMessages.length <= 5` (unread) to `allNewlyInserted.length <= 5` (all new) means 6+ total arrivals skip prefetch entirely, including the sent reply this PR is trying to surface. Defeats the primary motivation.

**3. Double `sync_complete` broadcast** — `_syncTick` already broadcasts `sync_complete`; adding a second one at the end of `prefetchNewMessageBodies` causes two full list reloads per sync cycle whenever new messages arrive.

Lower priority (can follow): `mergeServerThread` silently drops messages absent from the server response (archived on another device mid-sync), and the new `excludedPaths` construction adds O(N) extra DB queries per threaded list load.

maathimself:
Thanks for the contribution — went ahead and implemented this in-house given the bugs flagged last week. Closing in favor of the direct commit.

PR #163, opened 2026-06-16
fix: propagate updates to threadMessages and sync thread aggregate read state

### LINKED ISSUE #162: Read state not reflected in expanded thread sub-rows
## Describe the bug

When opening a message from inside an expanded thread (a sub-row in the conversation view), the sub-row stays visually unread — the bold "unread" styling and the unread dot don't go away. The API call to mark it read does fire and the change persists server-side, so refreshing the page shows the correct state. It's purely the in-page visual state of the sub-row that fails to update.

The top-level thread row in the list also doesn't update its unread count or read state when its last unread sub-message is opened, so an "all read" thread can still appear unread in the list until a manual refresh.

## Steps to reproduce

1. Use conversation view in a folder with an unread thread containing multiple messages
2. Expand the thread
3. Click an unread sub-row
4. The sub-row keeps the unread styling (bold, unread dot)
5. After opening every unread sub-message, the top-level thread row still shows an unread count

## Expected behavior

- Sub-rows in the expansion reflect their read state immediately on click
- The parent thread row's unread count and read state stay in sync with its sub-messages

saiththerobo:
## Summary

Fixes #162.

Conversation view rows render their unread state (dot + bold) from `unread_count`, but two paths left that value stale, so a thread could still look unread until a manual refresh:

1. **Expanded multi-message threads** — `updateMessage` in the store only touched the top-level `messages` array, so marking a sub-message read left its sub-row visually unread, and the parent row's `unread_count`/`is_read` never recomputed from its sub-messages.
2. **Single-message ("header only") threads** — `handleSelect` marked the clicked message read by setting only `is_read`. With no cached sub-messages, the store's aggregate resync had nothing to recompute from, so `unread_count` stayed stale and the row remained bold until a refresh.

## Changes

- `frontend/src/store/index.js`: `updateMessage` now applies the update across `messages`, `searchResults`, and `threadMessages`, then recomputes the parent thread row's `unread_count` and `is_read` from its sub-messages.
- `frontend/src/components/MessageList.jsx`: `handleSelect` passes `unread_count: 0` alongside `is_read: true` when marking a message read (and restores the prior value on the failure revert), matching the other mark-read paths. `handleSelect` only runs for single messages, so this is correct there; multi-message threads continue to clear via the store resync above.

## Testing

1. Use conversation view in a folder with unread threads.
2. Expand a multi-message thread and click an unread sub-row — bold/dot styling clears immediately; after opening every unread sub-message, the parent row's count clears and it de-bolds without a refresh.
3. Click a single-message (header-only) unread thread — the dot and bold clear immediately, no refresh needed.

---


maathimself:
Two regressions before this can merge:

**1. setMessagesReadState race** (line 568–569) — `updateMessage` now resyncs `unread_count` from `threadMessages` immediately, but `setCachedThreadRead` updates those sub-messages in a separate `set()` call on line 569. The resync fires first, reads stale sub-messages, and overwrites the `unread_count: 0` the caller just passed. Quick-action and context-menu mark-read on expanded threads leaves the row still bold.

**2. Keyboard paths skip `setCachedThreadRead`** — `markRead` (line 1207) and `onToggleRead` (line 1301) call `updateMessage` without ever calling `setCachedThreadRead`, so the resync reads stale sub-messages and immediately reverts the `is_read` change. j/k auto-mark-read and the u shortcut do not work for previously-expanded threads.

The sub-message case the PR targets is fixed correctly. The issue is that the resync also fires when updating a thread row directly, where `threadMessages` sub-messages are stale at that point. Simplest guard: skip the aggregate override when the caller already provided `unread_count`, and have the keyboard paths either pass `unread_count` explicitly or call `setCachedThreadRead`.

maathimself:
Thanks for the contribution — the approach here was right. I implemented a narrower fix in the store's updateMessage that guards the aggregate resync to sub-message updates only, avoiding regressions with keyboard paths and setMessagesReadState. Closing in favour of the in-house fix.

Issue #155, opened 2026-06-12
Run rules on existing emails (apply rules retroactively)


puqcloud:
**Is your feature request related to a problem? Please describe.**
When I create an inbox rule, it only applies to new incoming emails. There is no way to apply rules to emails that already exist in the mailbox. This is frustrating when setting up rules for the first time with thousands of existing emails.

**Describe the solution you'd like**
Add a "Run rules" button in the Rules settings panel that applies all existing rules retroactively to emails already in the mailbox.

**Describe alternatives you've considered**
Manually moving or archiving emails one by one, which is not practical with large mailboxes.

**Additional context**
This feature exists in other email clients — Gmail has "Apply to matching conversations", Thunderbird has "Run filters on folder". It would make the Rules feature much more useful on initial setup.

maathimself:
Implemented across three commits on main:
- 1f2a892 feat: run inbox rules retroactively on existing messages
- 684d423 fix: paginate all inbox messages in rules run, not just first 1000
- b879324 fix: raise proxy_read_timeout to 300s for rules/run endpoint

PR #164, opened 2026-06-17
chore: resolve all ESLint warnings and enforce zero-warning lint

saiththerobo:
## Summary

The backend and frontend ESLint configs were in place, but the lint step never actually enforced anything: `eslint src` exits 0 when only warnings are present, so warnings accumulated unnoticed (48 in the backend, 1 in the frontend). This PR clears every existing warning and then changes the lint command to fail on any warning, so the zero-warning state can't regress.

## Changes

- **Cleared all 49 ESLint warnings** across backend and frontend:
  - Dropped unused `catch` bindings to optional catch binding (`catch {}`).
  - Documented intentionally-empty `catch` blocks (required to satisfy `no-empty`; comment explains why the error is swallowed).
  - Removed dead variable initializers flagged by `no-useless-assignment` (`let x = null` → `let x`, where every read path reassigns first; the only reads use `!= null`, so `null` vs `undefined` is indistinguishable).
  - Removed the unused `TAG_BYTES` constant and the unused `validateHostLiteral` import.
  - Attached `{ cause }` to the re-thrown Microsoft token-validation error (`preserve-caught-error`), preserving the original error for diagnostics.
  - Added `// eslint-disable-next-line no-control-regex` only on the three regexes that intentionally match control characters (username / folder-name / filename sanitization).
- **Enforced a zero-warning ceiling**: both `lint` scripts are now `eslint src --max-warnings 0`, so CI fails on any warning. Rule severities in `eslint.config.js` are unchanged, keeping the warn/error gradient for editor feedback.

No runtime behavior changes.

## Testing

- `npm run lint` passes (exit 0) in both `backend/` and `frontend/`.
- `node --check src/index.js` passes (backend).
- Backend test suite: 224 passed.
- Frontend test suite: 814 passed.

---

### DIFF
--- a/backend/package.json
+++ b/backend/package.json
@@ -7,7 +7,7 @@
     "dev": "node --watch src/index.js",
     "test": "vitest run",
     "test:watch": "vitest",
-    "lint": "eslint src"
+    "lint": "eslint src --max-warnings 0"
   },
   "dependencies": {
     "archiver": "^7.0.1",
--- a/backend/src/routes/admin.js
+++ b/backend/src/routes/admin.js
@@ -249,7 +249,7 @@ router.post('/invites', async (req, res) => {
           });
           fromHeader = `${cfg.fromName || 'MailFlow'} <${cfg.fromEmail || cfg.user}>`;
         }
-      } catch {}
+      } catch { /* fall through to personal account */ }
     }
 
     // 2. Fall back to admin's first SMTP-enabled personal account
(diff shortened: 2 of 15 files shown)


maathimself (approved):
lgtm

maathimself:
Landed in e3ae6a3. Resolved the merge conflict by rebasing onto main before applying — the only conflict was in rules.js where the PR predated the #155 keyset pagination fix.

PR #231, opened 2026-07-10
fix: raise CardDAV parser entity cap for large address books

salmonumbrella:
## Summary

Large CardDAV address-book REPORTs can exceed fast-xml-parser's default entity-expansion limit. This change raises that limit to a generous finite cap and fails closed when a server reports a truncated snapshot.

## Changes

- configure the existing parser with a finite `10_000_000` total-expansion cap while preserving the previous depth setting
- reject top-level `507 Insufficient Storage` multistatus responses before extracting any cards
- retain the large-address-book regression and add coverage proving truncated results never return a successful subset

## Testing

- `npx vitest run` — 11 test files, 276 tests passed under Node 22.20.0
- `npm run lint`
- `node --check src/index.js`
- `git diff --check main...HEAD`

---

- [x] I have read and agree to the [Contributor License Agreement](https://github.com/maathimself/mailflow/blob/main/CLA.md).

### DIFF
--- a/backend/src/services/carddavClient.js
+++ b/backend/src/services/carddavClient.js
@@ -14,12 +14,9 @@ const parser = new XMLParser({
   ignoreAttributes: false,
   removeNSPrefix: true,   // <d:response> -> response, so parsing is namespace-agnostic
   trimValues: false,      // preserve vCard line structure inside <address-data>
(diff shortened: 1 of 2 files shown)


MikeScanlan5:
Thanks for opening this as part of my bug report. 🙏🏼🙏🏼

maathimself (approved):
lgtm

PR #243, opened 2026-07-12
fix: archive falls back to Gmail All Mail instead of silently no-oping

salmonumbrella:
## Summary

Gmail-style accounts have no `\Archive` folder; archiving there means removing the message from INBOX while it stays in All Mail. `resolveArchiveFolder` returned null for these accounts, so the archive route and inbox rules silently skipped the message.

## Changes

- Add an `\All` fallback in `resolveArchiveFolder` and an `isAllMailFolder` helper.
- At the archive call sites, remove the INBOX copy instead of duplicating the message into All Mail.

## Testing

`mailUtils.test.js` covers the fallback and `isAllMailFolder`; `inboxRules.test.js` gains an All-Mail archive case. The full backend suite passes.

---


maathimself (approved):
lgtm

PR #245, opened 2026-07-12
fix: warn when shortcut key assignments collide

salmonumbrella:
## Summary

`buildKeyMap` and `buildModKeyMap` let a later action silently take a key already bound to an earlier one, so a customized shortcut could disable a default with no trace.

## Changes

Both builders now `console.warn`, naming the two colliding actions and the winner.

## Testing

New `defaultShortcuts.test.js` covers both builders and the collision paths. The full frontend suite passes.

---


maathimself (approved):
lgtm

PR #249, opened 2026-07-12
fix: keep raw URLs and markup out of message preview snippets

### LINKED ISSUE #248: Message list previews show raw URLs and HTML markup instead of body text
Inbox previews often show junk instead of readable body text. Real examples from my instance:

- `( https://f90a918a.click.convertkit-mail2.com/8…`  newsletters whose text/plain part embeds the link URL in parentheses after the link text; the preview leads with the tracking URL.
- `<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1…` some senders put a raw HTML document in the text/plain part, and the preview shows the markup itself.
- `() Relink to We stopped importing transactions…`  leftover empty parentheses at the start of the preview.

Root cause: the snippet builder treats the text/plain part as ready-to-display prose, so link targets embedded by HTML-to-text converters and raw HTML shipped as text/plain pass straight into the stored snippet.

Expected: previews should behave like RFC 8970 (IMAP PREVIEW) previews an abbreviated plain-text representation of the message content with formatting markup removed, i.e. readable body text with link text preserved and no raw URLs or tags.

salmonumbrella:
## Summary

Message list previews showed raw link URLs (`( https://…click.convertkit-mail2.com/8… )`), HTML markup (`<!DOCTYPE html PUBLIC …`), and leftover empty parens instead of body text. This extends the canonical snippet builder in `messageParser.js` to drop link targets while keeping link text, route HTML documents shipped as text/plain through the existing HTML stripper, and sweep leftover empty brackets — the same approach as Nylas sync-engine snippets and RFC 8970 IMAP PREVIEW. Fixes #248

## Changes

- `snippetFromBody` detects HTML-shaped text/plain parts (document-level opener or dense closing tags) and routes them through `buildSnippetFromHtml`.
- Plain-text cleanup drops `label ( URL )`, `<URL>`, and bare `https://`/`mailto:`/`www.` link targets (link text survives; scheme-less bare domains like "example.com" in prose are kept), then collapses leftover empty `()`/`[]`/`<>` and standalone `**`/`__` runs.
- `parseMessage` (IMAP sync and background snippet indexer path) now delegates to the same canonical builders instead of a partial inline copy.
- Previously stored junk snippets are not retroactively regenerated: the background indexer only fills NULL/empty snippets, so old rows refresh only when a message body is fetched fresh from IMAP or the message re-enters the recent sync window.

## Testing

Added 13 vitest cases to `messageParser.headers.test.js`: the three real-world junk shapes above, HTML-in-plain routing, a sync-path `parseMessage` case, and do-not-over-strip regressions (parenthesized prose and times, bare domains, inline emphasis, prose mentioning an HTML tag). `npm test` passes 335/335 and `npm run lint` is clean in `backend/`.

---


maathimself (approved):
lgtm

Issue #222, opened 2026-07-08
Allow Message Rules to be Reordered


computrav:
**Is your feature request related to a problem? Please describe.**
The order of Rules cannot be changed, which prevents the full use of the "Stop processing further rules if this rule matches" option. I.e., if you create a new rule it goes to the bottom... but if this is something you want to run first and then stop processing further rules, there's no way to move it to the top.

**Describe the solution you'd like**
- Allow simple drag/drop to change the order of the rules (like the UI for re-ordering Favorited folders)

PR #253, opened 2026-07-12
feat: Add the ability to reorder filter rules

Lxtte1:
## Summary
This PR adds the frontend ability to reorder inbox rules.  It uses the existing backend API method `/rules/reorder`.  This request is in reference to #222.

## Changes
- Drag-and-drop reordering in settings
- A grab handle on each rule card

## Testing
I tested this by creating a bunch of rules, and moving them around a lot in different ways.  Nothing broke and performance was not degraded.

---


maathimself:
Hi @Lxtte1 , thanks for this.
Can you redo this with the native drag-and-drop the app already uses (see favDragIdx / reorderFavoriteFolders in Sidebar.jsx) instead of @dnd-kit? I'd rather not add the dependency when the codebase already does drag-to-reorder for favorites. Same /rules/reorder call, just native handlers.

Lxtte1:
It's gone now, sorry about that.

maathimself:
Not a problem,  just two things before merge:

1. `handleReorderRules` mutates the `rules` state array in place, then calls `setRules` with the same reference — so it only re-renders because the neighboring `setRuleDragIdx`/`setRuleDropIdx` calls happen to fire. Splice a copy instead:

   ```js
   const next = [...rules];
   const [moved] = next.splice(from, 1);
   next.splice(to, 0, moved);
   setRules(next);
   ```

   Also revert to the previous order if `api.reorderRules` fails (a failure currently leaves the UI reordered with no rollback), and drop the unused `updated`.

2. The `onTouch*` handlers (and the `onContextMenu` override) were copied from the favorites row but don't work here: they never start a drag, and `draggable` is off on mobile, so there's no touch reorder. Worse, the `preventDefault()` on touchstart suppresses tap-to-click on the row's checkbox/Edit/Delete buttons on touch devices (favorites compensates by handling taps in `onTouchEnd`; this row relies on `onClick`). Please remove them — desktop drag is enough for a settings panel.

Once those are in, good to go.

Lxtte1:
I've tried with a visual indicator that something is happening in case the server is lagging.  It just extends the opacity effect until it get an _"Okay"_ response.  I have the commit ready if that's something you want.

maathimself:
Let's skip the opacity indicator. Against a local backend the reorder is basically instant, and it adds a state that can get stuck. Simpler to make it optimistic:

1. In `handleReorderRules`, apply the order immediately and revert on failure:

   ```js
   function handleReorderRules(from, to) {
     const prev = rules;
     const reordered = [...rules];
     const [moved] = reordered.splice(from, 1);
     reordered.splice(to, 0, moved);
     setRules(reordered);
     api.reorderRules(reordered.map(r => r.id)).catch(() => setRules(prev));
   }
   ```

2. Delete the two leftover refs `ruleLongPressTimer` and `ruleTouchStart` — they're unused now and fail `npm run lint` (`--max-warnings 0`).

Then it's good to merge. Thanks!

Lxtte1:
I was hesitant to show a change that hasn't been confirmed.

maathimself (approved):
lgtm

PR #250, opened 2026-07-12
feat: opt-in GTD workflow with IMAP-folder-backed states

salmonumbrella:
## Summary

Opt-in, per-account GTD triage built on real IMAP folders — the five states (Todo, Watch, Delegated, Someday, Reference) are label folders that sync to every client and survive MailFlow itself. Off by default (Settings → Categories → GTD); accounts with GTD off behave exactly as before.

Reworked per your feedback, all three conditions addressed:

1. **petdex.dev dependency dropped entirely.** Nothing calls out to petdex — the fetch pipeline, its routes, UI, and locale strings are gone. The Inbox-Zero pet is the built-in SVG dog by default, plus an optional import-your-own path (pet.json + spritesheet uploaded directly, validated server-side by magic bytes with size caps; imported pets are private to their importer). `rg -i petdex` over the source returns nothing.
2. **Sidebar button restyle split out** — no longer in this branch; it now lives in #254.
3. **Rebased onto current main** (v2.5.0) — the conflicts with the recent merges are resolved, and the GTD theme honors the new `--accent-text` contract from #251.

**Stacked on #257 (generic right sidebar):** the first commit is that branch, the second is the GTD feature composing it. Until #257 merges this diff shows both; after it merges, a rebase drops the shared commit.

## Changes

- **Classify** from the keyboard (`t`/`w`/`d`, remappable), context menu, or hover actions. COPY semantics keep the INBOX message in place. **Done** strips the label(s), marks the whole thread read, and archives the INBOX copy.
- **Right sidebar sections** (first consumer of #257's shell): Todo → Waiting → Reference → Someday. Waiting merges Watch + Delegated with kind-colored aging chips (the older waiting date wins; red tint past 14 days) and a deduplicated unread rollup.
- **Message-list tab pills** per state, AI-generated gists for waiting heads (uses the existing per-user AI config; off when unconfigured), and per-account folder-name configuration with collision checks.
- **Opening an unread sidebar thread** applies the user's immediate/delay/manual mark-read preference to the whole thread, with identity-scoped timer ownership (a pending delayed read is cancelled if you explicitly act on that thread; unrelated triage doesn't touch it).
- **Live updates** over the existing WebSocket (`gtd_sections_updated`, debounced refetch, reconnect refresh for GTD users). Replying to a Todo/Someday thread auto-clears the label; sync ticks self-heal drift.
- **GTD theme** (dark, per-state accents), conforming to the theme var contract (new parity test pins it).
[...]

PR #254, opened 2026-07-13
fix(sidebar): hover state, tooltip, and secondary-gray tone for the collapse button

salmonumbrella:
## Summary

The sidebar button restyle split out of #250, as requested there. The sidebar hamburger now matches the app's other icon buttons instead of being a bare tertiary-gray glyph with no feedback. (This PR previously carried a hover-only version; it now contains the full restyle that was originally part of #250's diff.)

## Changes

- Transparent background with a bordered hover treatment (`--bg-tertiary` fill + `--border` outline on hover) and the shared `btn-press` affordance.
- Rest color moves from `--text-tertiary` to `--text-secondary`, matching sibling header icons.

## Testing

- Full frontend suite 1124/1124 (includes the locale key-parity suite covering the new key in all 7 locales), lint clean.
- Verified visually: hover shows the fill+border, tooltip renders, mobile close (×) path unchanged.

---

- [x] I have read and agree to the [Contributor License Agreement](https://github.com/maathimself/mailflow/blob/main/CLA.md).


maathimself:
The restyle itself looks good. Before I merge, can you drop the committed artifacts: the .superpowers/pr/sidebar-hamburger-hover.md file and the output/playwright/ directory. Those are local tooling and test output that shouldn't be in the repo. Once they're out and the branch is updated I'll get it in.

salmonumbrella:
Sorry about that, fixed!

maathimself (approved):

PR #256, opened 2026-07-13
feat: add Fastmail identity and Masked Email alias sync

salmonumbrella:
## Summary

Adds Fastmail as a provider: MailFlow syncs the account's sending identities and Masked Email addresses over the Fastmail API and lets you send from any of them.

## Changes

- Adds Fastmail in settings as a provider
- Adds Fastmail masked email address and alias support
- Works via a Fastmail API key (no OAuth needed) — write-only, encrypted at rest
- Synced identities appear in the compose From picker; replies come from the address that received the message, including `*@domain` catch-all identities
- The server authorizes every send against the synced identity list
- Search now matches recipients, so mail sent to an alias is findable

## Testing

- Backend: 523 tests (vitest); frontend: 1159 tests (node --test); lint and build clean on both
- Exercised against a live Fastmail account: token setup, identity and Masked Email sync, reply sender resolution, sending, recipient search

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -414,6 +414,23 @@ Gmail requires an **App Password** (not your normal password):
 | SMTP Port | `587` |
 | Username | your full iCloud email (`you@icloud.com`) |
 
+### Fastmail
+
+Fastmail uses two credentials for different responsibilities:
+
+1. Create a Fastmail app password and enter it in the account's **Password** field for IMAP and SMTP access.
+2. Create a Fastmail API token with **Email submission** and **Masked Email** permissions, then enter it in the **Fastmail API token** field.
+
+| Setting | Value |
+|---|---|
+| IMAP Host | `imap.fastmail.com` |
+| IMAP Port | `993` |
+| SMTP Host | `smtp.fastmail.com` |
+| SMTP Port | `465` |
+| Username | your full Fastmail email (`you@fastmail.com`) |
+
+MailFlow uses the API token to synchronize the sending identities and masked addresses that Fastmail authorizes. Manage provider-owned addresses in Fastmail, then use **Refresh addresses** in MailFlow when you need an immediate synchronization.
+
 ### Microsoft 365 / Outlook (OAuth2)
 
 **Work / school accounts (Microsoft 365):** require modern auth via Azure App Registration:
--- a/backend/migrations/0030_fastmail_alias_sync.sql
+++ b/backend/migrations/0030_fastmail_alias_sync.sql
@@ -1,0 +1,62 @@
+-- Fastmail alias/identity sync (v0.1)
+--
+-- Mirrors a Fastmail account's sending identities and Masked Email addresses into
+-- MailFlow so the user can send from any of them. All added columns are nullable or
+-- defaulted so existing rows stay valid.
+
+ALTER TABLE email_accounts
+  ADD COLUMN IF NOT EXISTS fastmail_api_token TEXT,
+  ADD COLUMN IF NOT EXISTS fastmail_last_sync TIMESTAMPTZ,
+  ADD COLUMN IF NOT EXISTS fastmail_sync_error TEXT;
+
+ALTER TABLE account_aliases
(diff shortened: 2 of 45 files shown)


maathimself:
Thank you for this. I'm not going to merge this one, it's about direction rather than quality.

Where I want to take MailFlow is general JMAP as a mail transport for self-hosted servers, with Stalwart as the first target. That's a different shape from what this PR does. This is a Fastmail-specific alias and Masked Email sync, built Fastmail-first: the origin is pinned to api.fastmail.com and the session load hard-requires the Masked Email capability, which is a Fastmail vendor extension a server like Stalwart doesn't implement. A general JMAP layer needs the inverse, host-configurable, capabilities optional, providers as profiles the way MailFlow already handles IMAP (google, apple, purelymail, generic). 

Either way, I appreciate the work here and hope you keep contributing.

salmonumbrella:
@maathimself what I tried to solve here is send/receive from masked email aliases. I was able to work around it by using a Fastmail API token. Any idea how to handle this so a user can `reply-as` a masked email out of Fastmail?

maathimself:
Most of this already works without the Identity API.

Receiving from masked emails works today over IMAP — they forward to the inbox, so MailFlow already sees them.

For sending, you don't need JMAP either. MailFlow already has per-account aliases, and `send.js` sends from an alias when you pass `aliasId`. So add the masked address as an alias on the Fastmail account, select it as the From, and MailFlow sends it over Fastmail's SMTP. Fastmail accepts a From that belongs to your account on authenticated submission (it may ask you to verify the address once, which is easy since the masked address receives mail).

salmonumbrella:
@maathimself it doesn't work right now without the Fastmail API key. when an email comes into `my_funky_alias@fastmail.com` this works. Then you press `r` to reply in MailFlow, the only option you have to reply is your primary fastmail account Pressing `r` should automatically set `my_funky_alias@fastmail.com` as the account you're replying from. 

I'm worried without a PR like this, it's impossible to get the benefit of using aliases in MailFlow.

maathimself:
Manual aliases do already show up in the reply From picker — if you add `my_funky_alias@fastmail.com` as an alias on the account (Settings → the account → aliases), it becomes selectable when you reply and MailFlow sends over Fastmail's SMTP, no API key involved. So sending from the alias works today; I think the "only option is the primary account" is because it was never added as an alias.

You're right about the real gap though: MailFlow doesn't yet auto-select the alias a message was *delivered to* when you hit reply, so right now you'd have to pick it from the dropdown each time.

salmonumbrella:
What if the solution was having MailFlow autoregister an a Fastmail alias when a user presses `r`? 

- receive email to `my_funky_alias@fastmail.com`
- press r
- check if `my_funky_alias@fastmail.com` exists in MailFlow—if not
- MailFlow create `my_funky_alias@fastmail.com` in settings
- MailFlow automatically sets `my_funk_alias@fastmail.com` as the `reply-as`

Could this pattern also work for folks who use aliases in Gmail?

🤔

maathimself:
I like the direction.

I don't like auto-creating a persistent alias, though. Two issues: it pollutes settings (every mailing list, one-off, plus-address, or catch-all you ever reply to would get silently saved as an alias), and "delivered to X" doesn't mean you're allowed to send as X, a list address delivers to your inbox but you can't send as the list. Cleaner to set the From transiently for that reply and leave saving an alias an explicit action.

On Gmail: it won't auto-work the same way. Fastmail accepts a From for any address your account owns, so the pattern is clean there. Gmail rejects or rewrites any From that isn't a pre-verified "Send mail as" identity, so you'd have to verify the alias in Gmail first — no auto-register. So this is clean for permissive providers like Fastmail, but not universal.

salmonumbrella:
Does this mean we need the Fastmail API key to do a check when the user clicks `reply` if they can permissions to `send as X` alias?

salmonumbrella:
@maathimself I made a few other PRs like #280 and #279 which handle this. can you please reconsider? there's no way to `send-as` a Fastmail email alias right now.

PR #257, opened 2026-07-13
feat: right sidebar with user-configurable IMAP label sections

salmonumbrella:
## Summary

Adds a collapsible right-sidebar column: in **Settings → Right sidebar**, each user picks any of their IMAP folders per account and drags them into order; those folders render as sidebar sections showing thread heads with unread/total counts.

**Off** by default. With no labels configured the column does not render at all and the server emits nothing extra, so existing users see no change.

## Changes

- `email_accounts.right_sidebar_labels` (JSONB, migration `0034`) is an ordered array of folder paths — array order *is* section order. Paths are sanitized on write (traversal, control chars, reserved system folders, length/count caps); rejected entries are surfaced back to the form.
- `GET /api/right-sidebar/sections` scopes accounts by `user_id` before any `accountId` filter, so a foreign id can only narrow to nothing. The section SQL first bounds to threads that appear in a configured folder, then rolls up **thread-level** unread: a thread is unread when *any* copy of it is (inbox copy unread ⇒ thread unread), and duplicate copies collapse into one row.
- Live refresh via a dedicated `right_sidebar_updated` broadcast — a label folder is never the *selected* folder, so the message list's own refresh never touches these sections. The route-side gate matches by `thread_key` as well as `message_id` (a reply landing in INBOX must refresh a section holding an older message of the same thread). Emitted from all mail mutation routes (incl. mark-all-read and folder delete/rename/empty), and from the sync paths (new mail, flag changes, backfill, reconciled deletes, snooze wakeup).
- Renaming a folder remaps any configured label to the new path and invalidates the per-account config cache.
- The drafts exclusion drops configured labels from its `%draft%` name-heuristic list, so a folder merely *named* like "Draft Proposals" still aggregates (real drafts folders are already rejected at write time).
- Sidebar rows reuse the shared row internals (`RowHoverActions`, `messageRowTypography`, `formatDate`) and the generic `RightSidebar` shell that landed with GTD (#250). When GTD is active for the current context it keeps the sidebar seam; label sections render otherwise — composing both into one column is a follow-up product decision.
- Row actions (open deep-link, read/unread, star, move, archive, delete, reply/reply-all/forward, create rule, block sender, copy subject/sender) apply optimistically and reconcile against a debounced refetch; failure paths schedule an authoritative refetch since the server may have committed before failing later work.

---

- Only the inbox is synced on a tick, so mail filtered server-side straight into a label folder appears when that folder is opened, backfilled, or the account reconnects. Generalising the sync tick has real IMAP-load consequences and deserves its own decision.
- In unified view, same-named folders from different accounts merge into one section with summed counts.

## Testing

- Backend: 661 tests + lint clean. Frontend: 1344 tests + lint clean + build.
- Full migration chain `0001–0034` applied against a throwaway Postgres 16; `right_sidebar_labels` lands as `jsonb NOT NULL DEFAULT '[]'`; re-running `0034` is a no-op.
- Independently reviewed twice (adversarial pass on the section SQL/broadcast gate/auth narrowing, plus a Codex review of the final diff); all findings fixed or documented above.
- Verified end-to-end in a browser against a GreenMail IMAP stack: configure → sections appear live without reload, deep-link open, hover mark-read with optimistic update + reconcile, unread rollups across folder copies.

## Contributor License Agreement

- [x] I have read and agree to the [Contributor License Agreement](https://github.com/maathimself/mailflow/CLA.md).


salmonumbrella:
@maathimself this was challenging to strip out, it might need extra attention from you.

maathimself:
Thank you but I'm going to decline this.

The core issue is that this ends up as a second sidebar system alongside the GTD sidebar, and the two are really the same primitive: folder-backed sections in the right rail with counts and row actions. That's why they can't share the column (GTD wins and the label sections silently don't render) and why so much of `RightSidebarLabels` mirrors `GtdSidebarContent`. It also overlaps with favorite folders, which already cover pinning and unread counts. So merging this means owning two parallel sidebars plus an unresolved decision about how they coexist.

If I do want configurable folder sections in MailFlow, I think the right shape is a single sections sidebar where GTD is just one preset arrangement, rather than two separate systems. That's a design I'd want to own and build deliberately, so I don't want to lock in the parallel version now.

Genuinely appreciate the effort here though.

PR #259, opened 2026-07-13
feat: name the collapsed sidebar rows on hover

salmonumbrella:
## Summary

Collapsed, the sidebar shows each account as a coloured initial and All Inboxes as a bare icon. Hovering tells you nothing. A collapsed account now names its address, and All Inboxes its label.

## Changes

- **`utils/sidebar.js`** (new). `collapsedTooltip` returns a label only while the rail is collapsed, since an expanded row already prints its own; it stays quiet for an account with no address rather than opening an empty tooltip. `activateOnKey` restores the Enter/Space activation a real `<button>` gives for free, swallowing Space so the page cannot scroll.
- **`Sidebar.jsx`**. The account row and `NavItem` draw `title`, `aria-label`, `role` and `tabIndex` from those helpers. The account row takes the button role only while collapsed: expanded, it holds the expand toggle, and a button cannot nest inside a button.

## Testing

- `utils/sidebar.test.js` (new): label only when collapsed, silence on a blank address, Enter and Space activate, Space is prevented, every other key passes through.
- `npm run lint` clean, `npm test` 1129/1129, `npm run build` succeeds.
- Not run against a live backend. The helpers are unit-tested; the wiring is attribute-level.

---


maathimself (approved):
looks good

PR #260, opened 2026-07-13
feat: show domain favicons as sender avatar fallback

salmonumbrella:
## Summary

Adds optional sender avatars → domain favicon → initial. The backend proxies favicons from Twenty's [favicon service](https://github.com/twentyhq/favicon) and caches them in Redis; the browser only talks to MailFlow.

## Changes

- Adds `GET /api/sender-favicons/:domain` — authenticated, rate-limited, preference-gated, `no-store`
- Sends the provider only the normalized sender domain, never local parts or user identifiers
- Accepts only square PNGs ≤64px and ≤64KB, over HTTPS with no redirects and a 5s timeout
- Resolves subdomains recursively to the registrable domain (`mail.notion.so` → `notion.so`), with a public-suffix stoplist and a 10-label cap
- Caches per domain in Redis: 7-day hits, 6-hour misses, 5-minute transient failures
- Adds a default-on "Sender favicons" toggle in Settings → Appearance; it fails closed before hydration and the backend enforces it independently
- Writes nothing to contacts or vCards; adds no dependencies or migrations

## Testing

- Backend: 442 tests (vitest); frontend: 1,143 tests (node --test); lint and build clean on both
- Exercised against a live account: fallback ordering, disabled-state 404s, subdomain walk, and cache persistence verified in the running app

---


maathimself:
Rebased onto main and landed (kept your authorship). Two changes on the way in: I set it off by default to match the existing Gravatar option, and folded it into the avatar chain so it composes with Gravatar (contact photo, then Gravatar, then domain favicon) instead of replacing it. Thanks @salmonumbrella!

PR #263, opened 2026-07-14
feat: bidirectional CardDAV contact sync

salmonumbrella:
## Summary

Makes CardDAV contact sync two-way: MailFlow already reads contacts from a CardDAV server (#209); this adds create, update, and delete back, with conflict handling.

Same posture as the read-only client: **standards only, no vendor pinning**. Nothing branches on a provider name or hostname, discovery is RFC 6764/6352, and every server capability is discovered and optional — a server that advertises nothing still works read-only. Verified against Nextcloud; the iCloud- and Fastmail-shaped fixtures serve as conformance cases, not provider profiles. [Stalwart](https://github.com/stalwartlabs/stalwart) therefore works out of the box.

## Changes

- **Remote-first mutations** — local state changes only after the server confirms; an ambiguous write persists its intent and recovers read-only rather than replay a `PUT`/`DELETE`.
- **Lossless vCards** — unmodeled properties (`X-*`, `CATEGORIES`, grouped `item1.*`, …) survive every edit path, including edits through MailFlow's own CardDAV server.
- **Conflicts** — simultaneous edits (and edit-vs-delete) get a side-by-side keep-mine / keep-theirs resolution.
- **Protocol** — RFC 6578 incremental sync with full-resync fallback, RFC 7232 conditional writes, honoured `Retry-After`.
- **UI** — contact editor with photos and typed fields, per-book read-only gating, a conflict queue, sync-state markers, all seven locales.
- **Migrations 0034–0037** add the sync schema; pre-existing data changes only via guarded duplicate-book consolidation and removal of the obsolete `dupMode` key.

## Testing

Backend 1,395 unit + 117 PostgreSQL-backed tests (real PostgreSQL 16, real HTTP CardDAV fixture server, wired into CI); frontend 1,418. Verified end-to-end in a browser against a disposable Nextcloud 34 and a live ~900-contact Nextcloud, where created contacts survive a full re-fetch with custom fields intact.

## Notes

1. **MailFlow's CardDAV server.** This PR prevents it from stripping retained properties on client `PUT`s — but should MailFlow run a CardDAV server at all, versus bridging to a real one through this client? Your call; the client stands alone either way.
2. **Multiple address books.** The client still pulls every book and writes into the first writable one. A follow-up giving each book a role (write target / subscribed / lookup-only, after JMAP's `isDefault`/`isSubscribed`) is ready — kept out to hold this PR to one capability.

---

### Contributor License Agreement

By submitting this pull request I confirm that:

[...]

PR #264, opened 2026-07-14
feat: per-book roles for CardDAV multi-address-book sync

salmonumbrella:
**Stacked on #263** (bidirectional CardDAV sync). Until that merges, GitHub shows both diffs. This PR adds the multi-book layer.

## Summary

One CardDAV credential often exposes several address books — a curated Apple Contacts book alongside the "Email contacts" book you want MailFlow writing to. The client pulled every book and wrote to the first writable: thousands of unwanted contacts in, writes to the wrong book. This PR assigns each book an explicit role and makes publication to a shared book deliberate.

## Changes

- **Per-book roles**, managed in the admin panel and enforced across sync, export, and conflict handling — modelled on JMAP's `isDefault`/`isSubscribed` split rather than a single flag:
  - *Write target* — exactly one; all MailFlow writes go there and nowhere else. A stable alias keeps the role when re-discovery changes the book's canonical URL.
  - *Subscribed* — pulled and shown in the contact list.
  - *Lookup source* — pulled into the sync ledger only, to resolve inbound senders (names, avatars) without materializing contacts.
  - Neither — ignored; its ledger rows are dropped.
- **Deliberate publication.** Editing an auto-collected contact no longer silently publishes it; an explicit promote action does. Sending mail to someone still marks them for autocomplete but no longer publishes them.
- **`publishEmailedContacts` setting, default off.** When off, nothing reaches shared books without deliberate action; when on, MailFlow fills books with correspondents. Pre-existing exports continue (the migration backfills intent). A separate `carddav_publish_intent` column keeps `send.js` untouched and preserves autocomplete ranking on `is_auto`.
- **Migrations 0038–0040**: role, alias, and publish-intent columns with behaviour-preserving backfills; the only pre-existing object touched is a widened mapping-status CHECK.

## Testing

Backend 1,466 unit + 171 PostgreSQL-backed tests; frontend 1,450. Tests verify both toggle states: when off, an emailed-only contact produces zero PUTs; when on, it exports to the write-target book.

## Note

macOS Contacts.app cannot handle multi-collection CardDAV accounts (Stalwart's maintainer hit the same limit). When MailFlow's CardDAV server exposes multiple books, it hits this client limitation — something to document, not fix server-side.

---


maathimself:

PR #278, opened 2026-07-16
feat: reply from the address a message was delivered to

salmonumbrella:
## Summary

When a message arrives at an alias through a BCC delivery or a catch-all, the alias never appears in To/Cc — so the existing reply alias auto-selection (which matches To/Cc/From against the account's aliases) misses, and the reply comes from the primary address. This follows up on the #256 discussion: "MailFlow doesn't yet auto-select the alias a message was delivered to when you hit reply."

This PR captures the envelope recipient (`Delivered-To` / `X-Delivered-To` / `X-Original-To` / `Envelope-To`) at ingest and lets the reply alias matcher check it first, ahead of To/Cc/From. Selection stays transient — nothing is auto-created or persisted into aliases, per the direction in #256. Existing messages carry no capture (NULL) and behave exactly as today, then backfill on their next sync touch; the send path is untouched — the matched alias flows through the existing `aliasId` mechanism.

## Changes

- Migration 0035 (IF NOT EXISTS-guarded, numbered after main's current 0034): nullable `messages.delivery_addresses JSONB`, same shape as `list_unsubscribe`
- `messageParser.js`: `parseDeliveryAddresses()` reads the four delivery headers from the already-parsed header map (reuses `parseMailboxList`; no IMAP fetch changes), deduped and capped at 50 addresses (sender-controlled input, capped like subject/snippet)
- `imapManager.js`: persists the parsed addresses at the message INSERT sites with the same `COALESCE` upsert plumbing as `list_unsubscribe` (the synthesized sent-copy record has no delivery headers and is deliberately left out)
- `messageService.js` + `mail.js`: expose `delivery_addresses` on the message shapes the frontend actually replies from (list, thread, single message, resolve)
- `MessagePane.jsx`: the inline reply-alias matcher is extracted to `frontend/src/utils/replyAlias.js` (pure, testable) and now checks delivery addresses first; the existing To/Cc/From matching semantics — including the alias-creation-order tiebreak for multi-alias accounts — are preserved bit-for-bit and pinned by tests

## Testing

- Backend: 633/633 (vitest), lint clean (`--max-warnings 0`) — new parser cases: repeated/multi-line headers, angle-bracket forms, cross-header dedupe, junk lines, the size cap
- Frontend: 1359/1359 (node --test), lint clean, production build clean — matcher tests pin: delivered-to beats To/Cc, the unchanged multi-alias tiebreak, malformed-JSON safety, case-insensitivity
- No new dependencies; no behavior change for existing messages (NULL capture ⇒ identical selection to today)

---

### DIFF
--- a/backend/migrations/0035_message_delivery_addresses.sql
+++ b/backend/migrations/0035_message_delivery_addresses.sql
@@ -1,0 +1,8 @@
+-- Addresses a message was delivered to (Delivered-To / X-Delivered-To /
+-- X-Original-To / Envelope-To), captured at ingest. Reply alias selection
+-- matches these in addition to To/Cc, so a message delivered to an alias
+-- via BCC or a catch-all still replies from that alias.
+-- Nullable like list_unsubscribe: NULL marks rows ingested before capture
+-- existed, so the sync upsert's COALESCE backfills them on a later re-sync.
+ALTER TABLE messages
+  ADD COLUMN IF NOT EXISTS delivery_addresses JSONB;
--- a/backend/src/routes/mail.js
+++ b/backend/src/routes/mail.js
@@ -140,7 +140,7 @@ router.get('/messages/:id', async (req, res) => {
              m.reply_to, m.in_reply_to,
              m.date, m.snippet, m.is_read, m.is_starred,
              m.has_attachments, m.account_id, m.category,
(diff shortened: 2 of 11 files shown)


maathimself:
Rebased onto main and squash-landed (kept your authorship). Renumbered the migration 0035 → 0037 to sit above main's current 0036. Thanks @salmonumbrella!

PR #279, opened 2026-07-16
feat: JMAP identity sync (host-configurable) with a Fastmail provider profile

salmonumbrella:
## Summary

Re-cut of #256 along the direction laid out there: host-configurable, capabilities optional, providers as profiles. A per-account, **read-only** JMAP identity sync answers one question — which addresses may this account send as — and stores the result in a private `sendable_addresses` table that is never rendered as a list anywhere (no settings pollution, no From-picker enumeration — the objections to the previous cut). On reply, when the delivered-to address is in that set, the From is set transiently for that one draft; at send time the server authorizes the explicit From against the same set. So "delivered to X" alone never grants "send as X": a mailing-list delivery resolves to nothing and replies from the primary, as today.

The JMAP client requires only `urn:ietf:params:jmap:core` + `urn:ietf:params:jmap:submission` (RFC 8620/8621 — Stalwart-compatible), takes its session URL from account config, and ignores capabilities it doesn't know (Masked Email included — split into a follow-up so this layer stays vendor-neutral). Fastmail is one provider profile: a settings preset (IMAP/SMTP hosts + JMAP session URL prefill) and an `imapManager` runtime profile. The scope is deliberately tiny — identity listing only, no JMAP mail transport — so it can fold into a future general JMAP layer.

Off unless a user configures a token; existing accounts and the existing send request shape are unchanged.

## Changes

- Migration 0036 (IF NOT EXISTS-guarded): `sendable_addresses` (private authorization set) + JMAP session URL / encrypted token / sync-status columns on `email_accounts`
- `jmapClient.js`: session load + `Identity/get` with response-shape validation; the session-advertised `apiUrl` is pinned same-origin to the session URL so the Bearer token can never travel to another host
- The user-configured session URL goes through the same SSRF guards as every other user-configured host: `validateHost` + `safeFetch` (per-hop IP validation), HTTPS required unless the host is genuinely private and the admin's connection policy allows private hosts — same rule as the CardDAV server URL
- `identitySync.js`: read-only reconcile (add/update/remove by provider id), including `*@domain` catch-all identities; runs on save and via a rate-limited manual refresh endpoint; no background scheduler
- `send.js`: optional `fromEmail`, mutually exclusive with `aliasId` (400 if both); authorized against the sendable set (exact, else wildcard cover) → 422 `SENDER_UNAVAILABLE` on a miss; the `aliasId` path is byte-for-byte untouched
- `GET /messages/:id/reply-sender`: resolves the transient reply From from the delivered-to capture (one-shot re-sync when stale); returns a single sender or null — never a list
- Compose: one pinned transient From option for that draft only; a 422 on send shows a toast and falls back to the account default
- Settings: JMAP session URL + API token fields (token write-only, encrypted at rest like `auth_pass`, surfaced only as a "configured" flag; clearing it wipes every synced row — in the same save as any other edited fields), plus last-sync/error status and a refresh button
- Fastmail provider profile: settings preset + `imapManager` runtime profile (`fastmail.com` / `messagingengine.com`)
- i18n: all 7 locales

## Testing

- Backend: 726/726 (vitest), lint clean (`--max-warnings 0`) — session/capability validation incl. a Stalwart-style custom host, apiUrl origin pinning, SSRF host-policy cases (private host blocked by default, plaintext-HTTP-to-public blocked even with private hosts allowed), reconcile add/update/remove/wipe, sender authorization (exact / wildcard / miss / header-injection), token clear + field update in one save, token never present in any response or error
- Frontend: 1381/1381 (node --test), lint clean, production build clean
- No new dependencies

Runtime-tested against a live Fastmail account on a real mailbox: token setup and validation, identity sync, delivered-to reply-From selection, and sending. Stacked on #278 (its diff is included here so the branch stands alone); I'll rebase this down to just its own changes once #278 lands.

---

### DIFF
--- a/backend/migrations/0035_message_delivery_addresses.sql
+++ b/backend/migrations/0035_message_delivery_addresses.sql
@@ -1,0 +1,8 @@
+-- Addresses a message was delivered to (Delivered-To / X-Delivered-To /
+-- X-Original-To / Envelope-To), captured at ingest. Reply alias selection
+-- matches these in addition to To/Cc, so a message delivered to an alias
+-- via BCC or a catch-all still replies from that alias.
+-- Nullable like list_unsubscribe: NULL marks rows ingested before capture
+-- existed, so the sync upsert's COALESCE backfills them on a later re-sync.
+ALTER TABLE messages
+  ADD COLUMN IF NOT EXISTS delivery_addresses JSONB;
--- a/backend/migrations/0036_jmap_identity_sync.sql
+++ b/backend/migrations/0036_jmap_identity_sync.sql
@@ -1,0 +1,34 @@
+-- Read-only JMAP identity sync (host-configurable — Fastmail is one provider
+-- profile, not a pinned default; Stalwart and other self-hosted JMAP servers
+-- work the same way). Only Identity/get is ever called; nothing is written
+-- back to the provider.
+--
+-- sendable_addresses is a private authorization set, never returned as a list
+-- by any API and never rendered in any UI list or From picker enumeration —
+-- "delivered to X" does not mean "allowed to send as X". It exists only so
+-- reply-time and send-time can transiently authorize a From that the
+-- provider has actually verified.
+ALTER TABLE email_accounts
+  ADD COLUMN IF NOT EXISTS jmap_session_url TEXT,
+  ADD COLUMN IF NOT EXISTS jmap_api_token TEXT,
+  ADD COLUMN IF NOT EXISTS jmap_identity_sync_at TIMESTAMPTZ,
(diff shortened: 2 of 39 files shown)


Clusters:
Mail-Server noob here.
I don't think my mail provider uses JMAP. It is a small German web hoster with an undisclosed mail server software stack.
I use well over 200 handles with my catch-all address, because I filter for spam with them.

Why can't it be a simple input field, pre-filled with the address from which it originated?
E.g.:
image

I know there are cases, where you don't want this behavior (e.g. for copy or redirect mail-accounts), therefore this should be a mail account boolean option to toggle?

maathimself:

PR #280, opened 2026-07-16
feat: sync Masked Email addresses when the server offers the capability

salmonumbrella:
## Summary

Follow-up to the JMAP identity sync: when a session advertises Fastmail's vendor extension (`https://www.fastmail.com/dev/maskedemail`) on a usable primary account, the sync also mirrors the account's enabled Masked Email addresses into the private sendable set. The capability is **detected, never required** — a server without it (Stalwart, any standard JMAP server) takes the identical code path with zero masked rows, and the vendor capability only enters the JMAP `using` array for the one call that needs it.

Result: mail arriving at a Masked Email address → reply → the From is already the masked address, transiently. Masked addresses appear nowhere else — no settings list, no picker enumeration, no alias creation, and nothing is ever written back to the server.

## Changes

- `jmapClient.js`: `sessionHasMaskedEmail()` capability check + `MaskedEmail/get` behind it, with the same response-shape validation and same-origin token pinning as `Identity/get`
- `identitySync.js`: enabled masks reconcile as `kind='masked'` rows in the same transaction as identities; a server that stops advertising the capability has its masked rows wiped on the next sync while identity rows are untouched
- `replySender.js` / `senderAuthorization.js`: dropped an incorrect `kind = 'identity'` filter so the sendable-set lookup is kind-agnostic (kind only records how a row was sourced, never whether it's usable) — with a seam-proof test driving a real send from a masked-sourced row with no other changes to the send path
- No frontend changes, no new strings, no new dependencies

## Testing

- Backend: 750/750 (vitest), lint clean (`--max-warnings 0`) — capability-absent path (never rejected, never requested), reconcile add/update/remove, capability-disappears wipe, disabled/pending/deleted masks excluded, malformed `MaskedEmail/get` rejected, tampered `apiUrl` never receives the token, end-to-end masked-row send
- Frontend: 1381/1381 (node --test), lint clean, production build clean — no functional diff, re-run for completeness

Runtime-tested against a live Fastmail account with Masked Email addresses on a real mailbox. Stacked on #278 and #279 (their diffs are included here so the branch stands alone); I'll rebase this down to just its own changes as they land.

---

### DIFF
--- a/backend/migrations/0035_message_delivery_addresses.sql
+++ b/backend/migrations/0035_message_delivery_addresses.sql
@@ -1,0 +1,8 @@
+-- Addresses a message was delivered to (Delivered-To / X-Delivered-To /
+-- X-Original-To / Envelope-To), captured at ingest. Reply alias selection
+-- matches these in addition to To/Cc, so a message delivered to an alias
+-- via BCC or a catch-all still replies from that alias.
+-- Nullable like list_unsubscribe: NULL marks rows ingested before capture
+-- existed, so the sync upsert's COALESCE backfills them on a later re-sync.
+ALTER TABLE messages
+  ADD COLUMN IF NOT EXISTS delivery_addresses JSONB;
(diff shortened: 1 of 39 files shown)


maathimself:

PR #284, opened 2026-07-17
fix: sync the first message into a previously-empty folder despite a stored modseq

salmonumbrella:
## Summary

A folder that seeds its CONDSTORE baseline while empty can never ingest its first message: with no local UID the new-mail phase is skipped, the delta path only updates flags, and the watermark still advances — so the miss becomes permanent. Hit in production on a GTD Watch folder (server count 1, zero cached rows). Separately, alias changes took up to the owner-cache TTL to count as "me", long enough for the GTD engine to strip Watch from a just-ingested self-sent message.

## Changes

- `planModseqSync` treats an empty local cache with a nonempty server mailbox as an incomplete cache and forces the metadata-capable full scan, regardless of modseq. The scan stays bounded; historical gaps remain backfill's job.
- Alias create/update/delete invalidate the GTD owner-address cache immediately.
- Corrected sync comments that claimed the UID phase "ALWAYS runs".

## Testing

- Unit tests pin the new `planModseqSync` states; an integration test drives the real `syncMessages` through the exact production state and fails pre-fix.
- Route tests cover invalidation on success/404/failure; GTD tests cover masked-alias sender keeping Watch vs external sender clearing it.
- Full backend suite + lint green on Node 22.

---

### DIFF
--- a/backend/src/routes/accounts.aliases.test.js
+++ b/backend/src/routes/accounts.aliases.test.js
@@ -1,0 +1,170 @@
+import { describe, it, expect, vi, beforeEach, beforeAll, afterAll } from 'vitest';
+
+// Alias CRUD is exercised through the mounted accounts router so these tests cover the
+// ownership checks, successful mutations, and the owner-address cache boundary together.
+// The DB, app entrypoint, and auth middleware are stubbed to keep the harness isolated.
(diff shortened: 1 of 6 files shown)


maathimself (approved):
lgtm

PR #285, opened 2026-07-17
fix: stop style/script contents, malformed tags, and divider runs leaking into snippets

salmonumbrella:
## Summary

Stored snippets could lead with raw markup: `<style>` contents survived tag stripping, a `>` inside a quoted attribute terminated the naive tag regex early (leaving fragments like `<p style=…>` in a quoted `src`, divider rules, degenerate text part with and without an HTML sibling), each verified to fail pre-fix.
- Prose guards pinned: braces like `{1, 2, 3}`, `a < b`, and lone `<b>` mentions stay untouched.
- Full backend suite + lint green on Node 22.

## Follow-up

- #287 fixes a pre-existing (on `main`) quadratic in the tag-strip scan — bodies with many bare `<` and no closing `>` stalled ingest for ~60s at 300KB — found while profiling this PR's snippet path. It is stacked on this branch.

---

### DIFF
--- a/backend/src/routes/mail.js
+++ b/backend/src/routes/mail.js
@@ -73,6 +73,10 @@ function snippetIsGarbled(s) {
     /&[a-z][a-z0-9]*;/i.test(s) ||   // undecoded HTML entity
     /##[^#]*##/.test(s) ||             // unexpanded template placeholder
     /-->/.test(s) ||                   // dangling HTML comment fragment
+    /\{[^}]*[:;][^}]*\}/.test(s) ||   // stored CSS rule block
+    /<[a-z][^>]*>/i.test(s) ||         // raw HTML tag
+    /<\/[a-z][a-z0-9:-]*\s*>/i.test(s) || // stray closing HTML tag
+    /([=_*#~-])\1{3,}/.test(s) ||      // decorative divider run
     /\[[^\]]+\]\(https?:\/\//.test(s)  // Markdown link syntax from ESP text/plain generators
   );
 }
--- a/backend/src/services/messageParser.js
+++ b/backend/src/services/messageParser.js
@@ -44,15 +44,26 @@ export function decodeNamedEntity(_, name) {
 }
 
 // Detect a text/plain part that is actually a raw HTML document — some senders
-// put the full HTML body in the text/plain alternative. A document-level opener
-// is definitive; otherwise require several closing tags near the start so prose
-// that merely mentions a tag ("use the <b> element") is not misrouted. Only the
-// head of the body is scanned to keep this cheap on large messages.
+// put the full HTML body in the text/plain alternative. A document-level opener,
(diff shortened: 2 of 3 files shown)


maathimself:
Thanks for this — the snippet-junk problem is real and most of the change is good. One blocker before merge.

**`snippetFromBody`'s CSS-block regex is quadratic:**

```js
.replace(/[^\s{}<>][^{}<>]*\{[^{}]*[:;][^{}]*\}/g, ' ')
```

`[^{}<>]*` re-scans from every position when there's no `{`, so it's O(L²) in the longest brace/angle-free run — and it runs synchronously on the full body at ingest/recompute. Measured on Node 22: real ~65KB newsletters take ~3s (vs ~3ms on main); 160k→15s, 320k→60s. Email content is attacker-controlled, so a crafted body is a trivial multi-minute freeze. `buildSnippetFromHtml` is fine; only this regex. Tests pass because every fixture is tiny.

Suggest gating on `text.includes('{')` (or bounding the pre-`{` scan) + a test that a ~200KB brace-free body finishes in <50ms. Happy to re-review after.

salmonumbrella:
Thanks for the review — fixed, and you were right about the measurements: I reproduced ~5s at 65KB and ~62s at 200KB on my machine with the old pattern.

The CSS-block strip now takes both of your suggestions plus one more layer, since a bare `includes('{')` gate alone still leaves a crafted opening (a single `{` followed by 300KB of `;` re-triggers quadratic backtracking through the declaration-content split):

- `includes('{')` gate — brace-free bodies skip the pattern entirely (~0.3ms at 200KB)
- lookbehind anchors candidate starts to run boundaries (the leftmost match always begins at one, so behavior is unchanged) instead of re-scanning from every position
- the pre-`{` selector scan is capped at 160 chars
- the separator check moved into a single forward-pass lookahead: `(?=[^{}]*[:;])[^{}]*\}` instead of the backtracking `[^{}]*[:;][^{}]*\}`

Output is byte-identical to the old regex on all fixtures plus 20k seeded fuzz cases over a brace/angle/separator-heavy alphabet. Crafted 300KB brace-bearing bodies (lone-brace, semicolon runs, tiny-word runs) all stay ≤ ~210ms worst case, linear in body size.

Tests added as requested: ~200KB brace-free body under 50ms, plus an adversarial brace-bearing case. Timing assertions take the fastest of a few runs so CI GC pauses can't flake them — a return to quadratic scanning overshoots the thresholds by orders of magnitude either way.

One heads-up while profiling: `buildSnippetFromHtml`'s tag-strip regex is quadratic on bodies with many bare `<` and no closing `>` (~60s at 300KB) — but that's pre-existing on `main` (`<[^>]+>` measures the same ~61s), not introduced here, so I left it out of this PR. Happy to send a follow-up for it.

maathimself (approved):
Looks good

PR #287, opened 2026-07-18
fix: linear-time tag stripping so bare-'<' bodies can't stall snippet ingest

salmonumbrella:
## Summary

Follow-up to #285, fixing the pre-existing quadratic flagged there while profiling: the tag-strip scan in `buildSnippetFromHtml` rescanned from every `<` toward a closing `>`, so a body with many bare less-than signs and none after them — crafted, or just unlucky prose in an HTML part — cost O(n²). A 300KB body took ~60s, synchronously, at ingest. This predates #285: `main`'s `<[^>]+>` measures the same ~61s on the same input.

## Changes

- The quote-aware tag regex admits exactly one parse at every character (a quote can only open an attribute value that runs to its matching quote), so where a tag closes is a pure function of the scan position. `stripTags` memoizes that walk in one right-to-left pass, then strips tags left-to-right with the regex's leftmost-match semantics.
- No caps or heuristics, so long-but-legitimate tags (inline styles, data-URI images) still strip instead of leaking into previews.

## Testing

- Byte-identical output to the old regex on 50k fuzzed markup samples over a `<>"'`-heavy alphabet, plus all existing fixtures.
- The crafted shapes (dangling opens, unpaired quotes, quoted-`>` runs) finish in single-digit milliseconds at 300KB; timing regression tests pin the envelopes using fastest-of-N runs so CI GC pauses can't flake them.
- Full backend suite green on Node 22 (655 tests).

> **Note:** stacked on #285 (both PRs edit the same function) — this PR shows two commits until #285 merges, after which only `b8ccbf9` remains to review.

---

### DIFF
--- a/backend/src/routes/mail.js
+++ b/backend/src/routes/mail.js
@@ -73,6 +73,10 @@ function snippetIsGarbled(s) {
     /&[a-z][a-z0-9]*;/i.test(s) ||   // undecoded HTML entity
     /##[^#]*##/.test(s) ||             // unexpanded template placeholder
     /-->/.test(s) ||                   // dangling HTML comment fragment
+    /\{[^}]*[:;][^}]*\}/.test(s) ||   // stored CSS rule block
+    /<[a-z][^>]*>/i.test(s) ||         // raw HTML tag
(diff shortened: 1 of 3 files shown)


maathimself:
Closing as I already pushed a fix, thank you.

PR #283, opened 2026-07-17
feat: semantic search via weighted FTS, body backfill, pgvector and MCP server

salmonumbrella:
## Summary

Overhauls search end to end: results ranked by relevance instead of date-only, bodies searchable across the whole mailbox, an optional embeddings pipeline adding a hybrid semantic mode, and a token-scoped MCP endpoint exposing search to agents. Everything costly or privacy-sensitive is off by default; stock Postgres still boots clean with semantic disabled.

## Changes

- **Weighted lexical ranking** — subject > sender > body (`setweight` A–D + `ts_rank_cd`), prefix matching as you type, served by a GIN index over a trigger-maintained `search_fts` column. Backfill runs as a resumable background drainer (fast-DDL migrations only, no boot-blocking rewrite); not-yet-backfilled rows fall back to the previous query path. Filter-only queries stay date-ordered; quoted phrases supported.
- **Body materialization** — a provider-gated background IMAP drainer fills `body_text` (apple/yahoo/generic on; Gmail/PurelyMail/Microsoft off given their throttle history), with circuit breakers, poison-message forward progress, and `background_jobs` progress rows.
- **Optional semantic search** — explicit opt-in, against any OpenAI-compatible `/v1/embeddings` endpoint (key encrypted at rest, host-validated, masked on read). Fingerprinted generations (config change ⇒ clean rebuild), a crash-safe fill worker, per-dimension HNSW partial indexes, re-embedding of late-arriving bodies. Query side is hybrid BM25∪ANN reciprocal-rank fusion behind an in-input Semantic toggle — lexical stays the default, with silent fallback while the index builds. Settings UI with explicit privacy copy and live Test/Build progress, in all 7 locales.
- **MCP server** — bearer-tokened Streamable-HTTP `/mcp` endpoint (SHA-256-hashed tokens minted in Profile, Origin-validated, rate-limited) exposing 12 tools including `semantic_search_messages` and `search_in_message`. Every call is scoped to the token owner's accounts; deletion is staged-only.

Evals on a real 18k-message mailbox: paraphrase Recall@20 goes 0% (current search) → 76% (hybrid); keyword MRR 0.43, the best of all modes tested.

## Deployment notes

- compose moves `postgres:16-alpine` → `pgvector/pgvector:pg16` (same PG16 major; the data dir mounts as-is). Run `REINDEX DATABASE` once after the first boot: the musl → glibc collation change can misorder existing text btree indexes (the app logs a loud warning when it detects this).
[...]

PR #294, opened 2026-07-19
feat(search): weighted FTS ranking + searchable message bodies

salmonumbrella:
Split 1/3 of #283, per the review request there. Keyword-search improvements only — no infrastructure changes, no new dependencies, runs on stock `postgres:16-alpine`.

## Summary

Search results are ranked by relevance instead of date-only, and message bodies become searchable across the whole mailbox.

- **Weighted lexical ranking** — subject > sender > body (`setweight` A–D + `ts_rank_cd`), prefix matching as you type, quoted phrases, served by a GIN index over a trigger-maintained `search_fts` column. Filter-only queries stay date-ordered.
- **Resumable backfill** — fast-DDL migrations 0041–0043 are crash-idempotent and boot-safe; `search_fts` fills via a background drainer, and not-yet-backfilled rows fall back to the previous query path. The legacy `search_vector` column stays until a follow-up removes it after real-install validation.
- **Body materialization** — a provider-gated background IMAP drainer fills `body_text` (Apple/Yahoo/generic on; Gmail/PurelyMail/Microsoft off given their throttle history), with pacing, poison-message forward progress, and per-account circuit breakers.
- The REST response carries a constant `mode: 'lexical'` field — the stable contract the semantic follow-up PR extends with real modes.

## Review fixes

- Rebased on current `main`; migrations are now 0041–0043.
- Connection refusals stop the body drainer immediately and arm its circuit breaker, including after partial batch progress.
- Searches without a selected account again respect `include_in_unified_inbox`.
- The initial FTS backfill count runs in a dedicated transaction with `SET LOCAL statement_timeout = 0`, so large mailboxes do not restart the same count every 30 seconds.

## Testing

- Local Node 22: backend 1,080 passed; frontend 1,553 passed; both linters clean; production build passed.
- Disposable exe.dev CrabBox, clean Ubuntu host with Node 22: same full test, lint, and production-build matrix passed.

---

### DIFF
--- a/backend/migrations/0041_search_fts.sql
+++ b/backend/migrations/0041_search_fts.sql
@@ -1,0 +1,55 @@
+-- Weighted full-text search column (search_fts) + version stamp, kept fresh by
+-- a BEFORE trigger. Fast, metadata-only DDL only (README D1): the nullable
+-- columns force no table rewrite on PG16, and the function/trigger are instant.
+-- Pre-existing rows are populated by the resumable drainer (ftsBackfill.js);
+-- the GIN index is built CONCURRENTLY in the separate no-transaction migration
+-- 0043 (a $$-quoted plpgsql body cannot survive the no-transaction ; splitter,
+-- so the trigger and the CONCURRENTLY index MUST live in different files).
+--
+-- The setweight(...) expression MUST stay identical to
+-- lexicalRepo.searchFtsExpr('NEW'); fts_version = 1 matches lexicalRepo.FTS_VERSION.
+-- Idempotent (IF NOT EXISTS / CREATE OR REPLACE / DROP IF EXISTS) because a
+-- crash before the schema_migrations INSERT retries this migration.
+
+ALTER TABLE messages ADD COLUMN IF NOT EXISTS search_fts tsvector;
+ALTER TABLE messages ADD COLUMN IF NOT EXISTS fts_version int;
+
+CREATE OR REPLACE FUNCTION messages_search_fts_refresh() RETURNS trigger AS $$
+BEGIN
+  -- Skip recompute on an UPDATE that changes none of the indexed source
+  -- columns (read/star flag flips, snippet-only writes), so the trigger does
+  -- not tax hot sync UPSERTs; this also lets the backfill's explicit SET win.
+  IF TG_OP = 'UPDATE'
+     AND NEW.subject      IS NOT DISTINCT FROM OLD.subject
+     AND NEW.from_name    IS NOT DISTINCT FROM OLD.from_name
+     AND NEW.from_email   IS NOT DISTINCT FROM OLD.from_email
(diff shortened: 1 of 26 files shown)


maathimself:
Thanks, this is great. The security side is clean too, I checked the FTS and SQL paths hard and there's no injection, and the auth and per-user scoping are right.

A few things before I can merge:

Blocking

1. Rebase and renumber the migrations. 0035/0036/0037 collide with main (0035_oidc_rp_initiated_logout, 0036_unified_inbox_accounts, 0037_message_delivery_addresses). Bump yours to 0040+.

2. The body drainer has no connection-refusal handling. The snippet indexer stops and backs off hard the moment a provider refuses a connection (isConnectionRefusal in imapManager.js) specifically so it doesn't pile on. bodyBackfill.js has nothing equivalent, so at an iCloud connection limit one batch fires ~100 fresh logins and the run retries to ~300 before giving up, which escalates the block and can starve the live IDLE/sync connection. This is a real problem I've hit with iCloud. Please add an isConnectionRefusal fast-stop, and make the breaker refusal-aware, right now it only backs off when a run makes zero progress. The description says "circuit breakers comparable to the snippet indexer" but that isn't the case yet.

3. Unified inbox scoping got dropped. searchService searches all enabled accounts and no longer honors include_in_unified_inbox, so an account a user deliberately excluded from the unified inbox now shows up in unified search. Please restore the include_in_unified_inbox filter for the unified (no specific account) case.

Should fix

4. On a large mailbox the ftsBackfill initial count(*) runs on the normal pool with the 30s statement_timeout, so it can throw and retry every 5 minutes forever, which means body search never starts on exactly the big mailboxes that need it. Run that count off the timeout-free path, or bound/drop it.

Everything else I found is minor and can come later. Once 1 through 3 are in I'll re-review.

salmonumbrella:
fixed the four review points in `ef15a6d`:

- rebased on current `main`; migrations are now 0041–0043
- connection refusals stop the body drainer immediately and arm the breaker, including after partial progress
- unified search again respects `include_in_unified_inbox`
- the initial FTS count uses a dedicated transaction with `SET LOCAL statement_timeout = 0`

verification:
- local Node 22: backend 1,080 tests, frontend 1,553 tests, both linters, frontend production build
- clean disposable exe.dev CrabBox: same full matrix passed

CI note: both GitHub jobs currently stop at the audit step because unchanged current-`main` lockfiles resolve `nanoid` 3.3.16 and the new advisory requires 3.3.17. This PR has no package or lockfile diff; the same `main` commit passed CI yesterday. I kept that dependency repair separate from this search PR.

maathimself:

maathimself:
Queued as #5 on #483's list. Closing until its turn, not on the merits: you did what #283's review asked, and this deserves a real review when it comes up. It will need a rebase by then.

PR #297, opened 2026-07-19
fix: honor user language for AI responses

sergeybychkovvvpgroup-beep:
Separate PR split out from #291 per maintainer review.

This keeps the IMAP MIME body-part fix focused and moves the unrelated AI language handling change into its own review.

Change:
- honor the user language setting for AI responses

### DIFF
--- a/backend/src/routes/ai.js
+++ b/backend/src/routes/ai.js
@@ -7,6 +7,21 @@ import { getConnectionPolicy } from '../services/connectionPolicy.js';
 
 const router = Router();
 
+const AI_LANGUAGE_NAMES = {
+  en: 'English',
+  ru: 'Russian',
+  de: 'German',
+  es: 'Spanish',
+  fr: 'French',
+  it: 'Italian',
+  zhCN: 'Simplified Chinese',
+};
+
+function aiLanguageInstruction(language) {
+  const name = AI_LANGUAGE_NAMES[language] || 'the user interface language';
+  return `Always respond in ${name}, unless the user explicitly asks for another language. For email drafting and rewriting, preserve the original email language when it differs from ${name}.`;
+}
+
 // ── Admin: AI provider configuration ──────────────────────────────────────────
 
 router.get('/admin/ai', requireAdmin, async (req, res) => {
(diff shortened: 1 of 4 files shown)


maathimself:
Thanks, but this branch was cut from #291's first commit rather than from main, so it carries the whole original IMAP/MIME change along with the AI one, its diff touches `imapManager.js`, `emailSanitizer.js`, and `mail.js`, and its commits include `353383c` "fix: fetch IMAP body parts individually".

That's the pre-review version, so merging this would re-introduce exactly what we removed in #291: `repairResidualQuotedPrintableHtml` (corrupts ~56% of real bodies and writes the corruption back to the DB cache), the un-guarded blanket per-part re-fetch, and `unwrapEmbeddedMimeText` with no recursion bound.

Could you re-cut this off main so it contains only the `ai.js` change? Since that change is entirely in one file, the simplest path is:

    git checkout -B fix/ai-language origin/main
    git checkout 9a6d0a7 -- backend/src/routes/ai.js
    git commit -m "fix: honor user language for AI responses"
    git push --force-with-lease

The AI-language change itself looks good, it just needs to stand alone. Thanks!

maathimself:
Following up here, since this still carries the #291 IMAP/MIME changes (`imapManager.js`, `emailSanitizer.js`, `mail.js`) on top of the `ai.js` change. The re-cut from main is still the blocker; the commands in my previous comment should do it.

Two small things to fold in once it stands alone, so the AI change is complete:

- Add a test for the language mapping (`aiLanguageInstruction` and its allowlist plus fallback). It is a small pure function and easy to cover.
- Decide whether the GTD "waiting" gist summaries (`gtdGist.js`) should also honor the user's language. Right now `/ai/chat` gets the language instruction but the gist path does not, so those summaries stay in the email's language. Either behavior is defensible, but it is worth a deliberate call rather than an accidental inconsistency.

The `ai.js` change itself looks good and is safe: the language value goes through the allowlist and is never interpolated raw into the prompt.

salmonumbrella:
Re-cut from current `main` is available as #329. It is one focused commit, contains only `backend/src/routes/ai.js` plus its new mapping/fallback test, and carries none of the IMAP/MIME files from `353383c`.

The original contributor fork rejects writes from the authorized `salmonumbrella` identity, so #297 itself could not be rewritten in place. #329 is the replacement branch.

GTD waiting gists intentionally continue using the source email language. They are cached content summaries rather than interactive UI responses; making them follow a UI-locale switch would require separate preference plumbing and cache invalidation.

maathimself:
Thanks for this. The AI language part is good to go. Could you drop the imapManager/MIME changes from this PR? That fix is already covered in #291, so keeping both here would double them up. If you trim this down to just the ai.js language change I'll get it merged. Thanks!

salmonumbrella:
I cannot rewrite #297 in place because its head branch belongs to the original contributor fork and rejects writes from my account. #329 is the clean replacement for merge; #297 can be closed as superseded.

maathimself:
Superseded by #329, which is the focused ai.js-only version and is now merged. Thanks @sergeybychkovvvpgroup-beep for the original - the MIME half went in separately as #291.

PR #298, opened 2026-07-19
fix: attached HTML/text files vanish from the attachment list

Monkey7539:
## Problem

Emails with an attached `.html` (or `.txt`) file show the paperclip indicator, but the attachment never appears in the attachment list — there is nothing to view or download. Worse, an attached HTML file can end up rendered as the message body, shadowing the real content.

## Root cause

`walkStructure` in `backend/src/services/imapManager.js` classifies MIME parts by **type before disposition**. An attached HTML file is `Content-Type: text/html` + `Content-Disposition: attachment; filename="..."`, so it hits the `text/html` branch and is pushed into `textParts` (message body) — it never reaches the attachment branch. The same happens to attached `text/plain` files.

Meanwhile `detectAttachments` in `messageParser.js` (which drives `has_attachments` / the 📎 indicator) keys on `disposition === 'attachment'` — so the indicator and the attachment list disagree for exactly these parts.

## Fix

Route any part with `Content-Disposition: attachment` to the attachment list **before** the text-type branches, mirroring `detectAttachments` so the two can never disagree. Behavior is otherwise unchanged:

- undisposed `text/html` / `text/plain` parts still flow to the message body (`multipart/alternative` untouched)
- `cid:`-referenced images stay inline
- named non-text parts without a disposition are still collected as attachments

## Tests

New `walkStructure` describe block in `imapManager.test.js`:
- attached HTML file → attachment list, not body
- attached text file → attachment list
- undisposed HTML in `multipart/alternative` → still body
- attachment-disposed image → attachment; `cid:` image → inline
- named PDF without disposition → attachment

Full backend suite passes (691 tests) and eslint is clean.

---

### DIFF
--- a/backend/src/services/imapManager.js
+++ b/backend/src/services/imapManager.js
@@ -335,7 +335,7 @@ function decodeAttachmentBuffer(buf, encoding) {
   return Buffer.isBuffer(buf) ? buf : Buffer.from(buf);
 }
 
-function walkStructure(node, results) {
+export function walkStructure(node, results) {
   if (!node) return;
   const type = (node.type || '').toLowerCase();
   if (node.childNodes && node.childNodes.length > 0) {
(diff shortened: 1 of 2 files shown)


maathimself:
Thanks @Monkey7539! I rebased this onto `main` (resolving a one-line import-list conflict with the concurrent PurelyMail IDLE change) and landed it as e86d177, keeping you as the commit author. Full backend suite (695 tests) and eslint pass clean. It went in as a rebased commit directly on main rather than through this branch, so I'm closing the PR out here. Appreciate the fix.

PR #319, opened 2026-07-28
fix: fail over mail server addresses

### LINKED ISSUE #318: Mail server connections do not fail over resolved addresses
**Describe the bug**

MailFlow resolves an IMAP/SMTP hostname to all A/AAAA records for validation, but `resolveForConnection()` returns only the first address. Every connection is then pinned to that single IP. If a provider advertises multiple addresses and the first one is unreachable from the MailFlow host, the connection times out even when another advertised address is healthy.

SMTP connection attempts can also outlive the frontend proxy timeout. Nodemailer's default connection timeout is about 120 seconds while the bundled nginx proxy waits 60 seconds, so the user sees a non-JSON `504 Gateway Timeout` / `Request failed` before the backend eventually logs `Send failed: Connection timeout`.

**To Reproduce**

1. Configure an IMAP account whose hostname resolves to at least two public IP addresses.
2. From the MailFlow host, make the first resolved address unreachable while leaving the second address reachable.
3. Connect or wait for a staleness probe.
4. Observe an IMAP connection/staleness timeout; MailFlow never attempts the healthy address.
5. For SMTP, make the resolved endpoint silently drop new connections and send a message.
6. Observe nginx return 504 at 60 seconds while the backend logs the connection timeout roughly a minute later.

A real deployment reproduced the IMAP case with a provider returning two IPv4 addresses: TLS to one consistently succeeded while the other consistently timed out, and DNS response order changed between queries.

**Expected behavior**

- Validate every resolved address before it can be used.
- Try the remaining validated addresses when an IMAP connection fails during establishment.
- Preserve the original hostname as TLS SNI/servername for every candidate.
- Bound SMTP connection and greeting timeouts below the frontend proxy timeout.
- Retry SMTP only for unambiguous pre-delivery connection failures so failover cannot duplicate a message.
- Prefer a recently successful address or temporarily de-prioritize failed addresses to avoid repeatedly waiting on a known-bad endpoint.

**Additional context**

`backend/src/services/hostValidation.js` currently builds `all = [...v4, ...v6]` and returns `host: all[0]`. This protects against DNS rebinding but prevents address fallback. IMAP and SMTP consumers receive only that pinned address.

The fix should retain the SSRF guarantees: all candidate addresses must be validated up front, and no later hostname lookup should occur between validation and connection.

**Environment**

- MailFlow: latest `main` / 2.7.1-based custom deployment
- Deployment: Docker Compose with bundled nginx frontend
- Backend: Node 22

YunQue0912:
## Summary

Fixes #318.

Mail server hostnames can resolve to several valid addresses, but MailFlow previously pinned every connection to the first one. An unreachable first address therefore blocked IMAP and SMTP even when another validated address was healthy. SMTP's default connection timeout could also outlive the frontend proxy timeout.

## Changes

- Preserve every validated A/AAAA address while keeping DNS rebinding protection: connection code receives only the addresses validated up front.
- Let Node try the remaining validated IMAP addresses while retaining the original hostname for TLS SNI and certificate verification.
- Add one SMTP transport wrapper for user mail, system mail, invite/password-reset mail, and SMTP verification.
- Limit each SMTP connection and greeting stage to 10 seconds and the failover sequence to a 45-second budget.
- Retry SMTP only for Nodemailer `CONN` failures, before authentication or envelope/data commands, so failover cannot duplicate delivered messages.

## Testing

- Backend lint: `npm run lint`
- Backend syntax: `node --check src/index.js`
- Backend tests: `710` passed across `35` files
- Backend production audit: `0` vulnerabilities
- Frontend lint: `npm run lint`
- Frontend build: `npm run build`
- Frontend tests: `1401` passed across `48` suites
- Frontend production audit at `audit-level=high`: passed (two existing moderate React Router advisories have no available fix)
- Added a local socket test proving same-family IPv4 fallback reaches the second candidate without another DNS lookup
- `git diff --check`

---

### DIFF
--- a/backend/src/routes/admin.js
+++ b/backend/src/routes/admin.js
@@ -1,10 +1,10 @@
 import { Router } from 'express';
 import crypto from 'crypto';
-import nodemailer from 'nodemailer';
 import { query } from '../services/db.js';
 import { requireAdmin } from '../middleware/auth.js';
 import { decrypt, encrypt } from '../services/encryption.js';
 import { validateHost, resolveForConnection } from '../services/hostValidation.js';
+import { createSmtpTransport } from '../services/smtpTransport.js';
 import { getConnectionPolicy, invalidateConnectionPolicyCache } from '../services/connectionPolicy.js';
 import { reloadAuthSettings } from '../services/authLimiter.js';
 import { imapManager } from '../index.js';
@@ -291,8 +291,7 @@ router.post('/invites', async (req, res) => {
           const sysResolved = await resolveForConnection(cfg.host);
           const sysTls = { rejectUnauthorized: true };
           if (sysResolved.servername) sysTls.servername = sysResolved.servername;
-          transport = nodemailer.createTransport({
-            host: sysResolved.host,
+          transport = createSmtpTransport(sysResolved, {
             port: cfg.port || 587,
             secure: (cfg.port || 587) === 465,
             auth: { user: cfg.user, pass },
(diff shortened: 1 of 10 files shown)


maathimself (approved):
lgtm

Issue #304, opened 2026-07-20
Archive should remove messages immediately during rapid and bulk keyboard triage


salmonumbrella:
## Problem

Archiving with the `e` shortcut does not keep up with fast, Superhuman-style inbox triage. Messages lag before disappearing, and some remain visible after the archive operation finishes.

I can reproduce this reliably:

1. Open a mailbox with at least ~40 messages.
2. Select/highlight roughly 40 messages.
3. Press `e` to archive the selection.
4. The UI stalls for about 3 seconds, then the rows still do not all disappear cleanly.

The same roughness is noticeable when moving through messages and pressing `e` quickly. It breaks the keyboard-first flow because the list does not immediately reflect the action.

## Prior attempt

I tried to improve archive triage in #247 by advancing selection before removing the archived row. That PR merged, but it did not solve this lag/stale-row behavior—especially for rapid or bulk archive.

## Expected behavior

- Remove every targeted row from the visible list immediately, before network/IMAP work completes.
- Advance/clear selection deterministically in the same local state update.
- Send one batch operation for a multi-selection rather than serial per-message work where possible.
- Reconcile in the background; on failure, restore the affected rows and show an error.
- Keep all cached list variants/counts consistent so a refresh or websocket event cannot briefly resurrect archived rows.
- The interaction should stay responsive with ~40 selected messages and with repeated `e` presses.

A short horizontal slide-out could be a nice finishing touch for the hotkey workflow, similar to Superhuman, provided it stays compositor-friendly and does not delay the optimistic removal.

## Useful references

- [Aster Mail bulk archive](https://github.com/Aster-Privacy/Aster-Mail/blob/main/src/hooks/use_email_list_bulk.ts#L215-L318): removes the selected IDs from local state up front, updates counts/cache, makes one batch archive call, and restores the snapshot on failure.
- [Aster Mail single-message archive](https://github.com/Aster-Privacy/Aster-Mail/blob/main/src/hooks/use_email_list_actions.ts#L292-L351): the same immediate-remove/rollback pattern for keyboard-style one-at-a-time triage.
- [OpenCo optimistic update commit](https://github.com/theopenco/llmchat/commit/56609e33ca942ab5f6c70478432a9b74082896fb): snapshots all matching cached list variants, drops archived rows immediately, rolls back on failure, and includes tests for multi-variant consistency.
[...]

PR #322, opened 2026-07-29
fix: prevent GTD done rows from resurfacing

salmonumbrella:
## Summary

Extends the instant, resurrection-proof archive behavior from `f7166e1` to the separate GTD sidebar `useGtdTriage` / `doneRow` path.

This is a follow-up to #304: clicking the GTD sidebar checkmark now removes the row immediately and keeps stale websocket/refetch results from bringing it back.

## Changes

- Guard state-scoped GTD removals while the request is pending and through a short post-success reconciliation window.
- Apply the removal guard at the GTD sections fetch boundary so refreshed data cannot resurrect an archived row.
- Capture a row-level snapshot and restore only that row on failure, preserving other rapid or concurrent triage actions.
- Add focused tests for immediate removal, pending/completed reconciliation, state-scoped rapid actions, partial archive failure, and safe rollback.

## Testing

- `cd frontend && npm test` — 1,414 tests passed
- `cd frontend && npm run lint`
- `cd frontend && npm run build`

---

### DIFF
--- a/frontend/src/hooks/useGtdTriage.js
+++ b/frontend/src/hooks/useGtdTriage.js
@@ -8,6 +8,7 @@ import {
 } from '../utils/gtd.js';
 import { openReplyFromMessage, openForwardFromMessage } from '../utils/composeFromMessage.js';
 import { resolveContextMenuMessage } from '../utils/contextMenuPolicy.js';
+import { doneGtdRow } from '../utils/gtdDone.js';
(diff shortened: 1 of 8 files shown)


maathimself:
Rebased onto main and squash-landed as 924fc88f8 (kept your authorship). Thanks @salmonumbrella!

Issue #219, opened 2026-07-08
Allow messages to be opened in new window


computrav:
**Is your feature request related to a problem? Please describe.**
Currently, you can only view 1 email at a time (unless you open a completely new Mailflow window of course).

**Describe the solution you'd like**
- Double-click on a message in the message list opens the email in a new window (probably not an actual window but a modal within the SPA). The UI is essentially exactly the same as the message pane that exists now, but is in a separate window.
- Replying/forwarding goes to the compose view, still in the same window
- "Closing tasks" like archiving, trash, moving, marking as spam closes the window
- Other tasks do NOT close window, such as Mark Read/Unread, Star, AI stuff, print
- The modal can be minimized, resized, closed (so you could for example open a few emails in diff modals, and minimize all of them, and be able to restore any of them -- so when you have 2+ modals minimized, they should get organized nicely and not just stack directly on top of each other)

maathimself:
Implemented in 54da62d. Double-click a message row (or right-click > Open in new window) to open it in a draggable/resizable/minimizable in-app window; multiple minimized windows dock side-by-side. Closing actions (archive/trash/move/spam/snooze) dismiss the window, other actions leave it open. Desktop-only.

PR #362, opened 2026-08-09
fix: stabilize threaded inbox triage

### LINKED ISSUE #361: Threaded Inbox rows expand on open and reappear during rapid archive
## The situation

With conversation threading enabled in the Safari installed web app, clicking a collapsed Inbox thread opens the newest message and expands every child row at the same time. Rapidly pressing `e` can also make an archived conversation disappear and then come back under an older message.

The two symptoms share the same thread-boundary problem: the row click owns both selection and disclosure, while keyboard archive and its optimistic guard operate on the current head UUID instead of the conversation's stable identity.

The count/chevron can collapse an expanded thread manually, and the existing archive request guard prevents an exact archived message UUID from returning during an overlapping list refresh. That is enough for flat rows, but not for a grouped conversation: once the head is archived, a sibling gets a different UUID and becomes the replacement head.

Archiving every folder copy would avoid that replacement, but it would also move Sent replies and other copies that are deliberately outside the active Inbox view.

## Ask

- Keep a collapsed thread collapsed when its row is opened; show the newest message in the reading pane.
- Make the count/chevron the only control that expands or collapses child rows, including in the Safari touch path.
- Archive every member of the conversation in the active folder and account scope while preserving Sent and other-folder copies.
- Guard the stable thread identity during pending/completed refresh windows, and recover cleanly from partial archive failures without stale cached children.

This is a follow-up to the exact-message refresh protection added in #267; flat-message behavior and backend archive semantics should remain unchanged.

salmonumbrella:
## Summary

Threaded Inbox triage had two coupled failures: opening a collapsed conversation expanded every child row, and rapid archives could let an older sibling return as the replacement head. This keeps selection, disclosure, and archive state on the right boundaries.

Closes #361.

## Changes

- Keep a thread collapsed when its row is opened; the reading pane still opens the newest message, and only the accessible count/chevron toggles child rows.
- Keep Safari touch handling from routing a count-button tap through the row-open path.
- Archive every conversation member in the active folder and account from keyboard, swipe, multi-select, and context-menu actions, while preserving Sent and other-folder copies.
- Guard pending and completed archive state by stable thread identity, scoped to the source folder and account, so a replacement head cannot reappear during a refetch.
- Use one undo transaction window, reject late Undo actions after commit starts, and keep rollback scoped to the action-time list view.
- Resolve destructive thread targets from a fresh server snapshot, bound concurrent thread loads, and chunk archive requests at the backend's 500-ID limit.
- Invalidate expanded-thread caches across partial failures and stale in-flight loads, while reconciling unread counts per account.

## Testing

- `cd frontend && npm ci`
- `npm test` — 1,579 passed
- `npm run lint`
- `npm run build`
- Repeated the full test, lint, and build gate on a clean remote Linux runner with Node 22.
- Reproduced the original selection behavior in the Safari installed web app.
- Verified the rebuilt local UI keeps row clicks collapsed, expands only from the count control, and refetches children after archive-failure recovery.

---

### DIFF
--- a/frontend/src/components/MessageList.jsx
+++ b/frontend/src/components/MessageList.jsx
@@ -21,7 +21,15 @@ import SenderAvatarImage from './SenderAvatarImage.jsx';
 import { shortcutBus } from '../utils/shortcutBus.js';
 import { createLatestRequest } from '../utils/latestRequest.js';
 import { pendingMarkReadMap, completedMarkReadMap, setPending } from '../utils/pendingReads.js';
-import { applyDeleteGuard, clearDeleteGuard, clearPendingDelete, setCompletedDelete, setPendingDelete } from '../utils/pendingDeletes.js';
+import { applyDeleteGuard, clearDeleteGuard, clearPendingDelete, setCompletedDelete, setPendingDelete, threadDeleteGuardKey } from '../utils/pendingDeletes.js';
+import {
+  archiveTargetsForFolder,
+  currentThreadLoadVersion,
+  findVisibleArchiveMessage,
+  invalidateThreadLoad,
+  isCurrentThreadLoad,
+  unreadCountsByAccount,
+} from '../utils/threadedArchive.js';
 
 // Folder icon for move picker
 function FolderIcon({ specialUse, size = 13 }) {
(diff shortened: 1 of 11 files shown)


salmonumbrella:
Follow-up commit `a5ec6be` now applies the thread-scoped archive transaction to keyboard, swipe, multi-select, and context-menu actions.

The reported `Unknown` rows are not a separate sender-parsing defect. They are `ThreadRow`'s `common.unknown` fallback rendering a resurrected thread head that has no sender fields. Preventing the stale sibling head from returning removes that symptom at its source.

maathimself:
Rebased onto current `main` (this branch had drifted 48 commits behind and conflicted with the #219 detached-window, GTD plugin-extraction, and list-selection refactors) and merged via the temporary `fix/threaded-inbox-triage-rebased` branch. Both your commits landed with authorship intact (8698a7f, bd57b3a), plus a small follow-up (92b0417) that accents the thread count pill, enlarges its mobile tap target, and fixes a pre-existing mobile thread-row highlight persisting after swipe-back. Thanks @salmonumbrella — closing as merged-by-rebase. Fixes #361.

PR #333, opened 2026-07-31
feat: undo inbox triage actions

salmonumbrella:
## Summary

Add a configurable `Cmd+Z`/`Ctrl+Z` Inbox triage undo shortcut while preserving native text undo in composers and editable controls.

## Changes

- Route the standard `undo` shortcut through the existing guarded shortcut dispatcher and limit it to the active Inbox message list.
- Give keyboard, context-menu, swipe, and bulk triage actions one atomic 4.5-second undo window with newest-action-first ordering.
- Share the delayed archive path across keyboard and pointer actions while preserving next-message selection and optimistic counters.
- Make newly confirmed GTD classifications reversible without removing pre-existing labels or changing COPY semantics.
- Add regression coverage for editable-target routing, rapid actions, exact expiry boundaries, non-UIDPLUS classification, and failure reconciliation.

## Testing

- Frontend: `npm test` — 1,579 tests passed.
- Frontend: `npm run lint` — passed with zero warnings.
- Frontend: `npm run build` — passed (existing chunk-size advisory only).
- Backend: `npx vitest run` — 990 tests passed.
- Backend: `npm run lint` — passed with zero warnings.

---

### Contributor License Agreement

By submitting this pull request I confirm that:

- [x] I have read and agree to the [Contributor License Agreement](../CLA.md).
- [x] My contribution is my own original work (or I have identified any
      third-party material and confirmed it is compatible with the CLA).
- [x] I have the right to submit this contribution under the terms of the CLA.

Issue #368, opened 2026-08-13
GTD is now a plugin, read before re-opening GTD work


maathimself:
@salmonumbrella, before you jump back into GTD, heads up that I moved it into its own plugin. Wanted to give you the rundown.

Short version: GTD is now self contained, so you can own it and iterate without me having to review every change for whether it could break core.

Where it lives now:
- Backend: `backend/src/plugins/gtd/`
- Frontend: `frontend/src/plugins/gtd/` (a few UI files are still under `components/` and `utils/` for now, they'll move later)

Main thing to know on the backend: GTD only talks to core through `backend/src/plugins/api.js`. That's the whole toolkit (labels, storage, summarize, mail/account reads, per account config, broadcast, etc). It can't import core services, the DB, or the mail engine directly. There's a lint rule that enforces this in CI (`npm run lint:plugins`), and at runtime it only gets a locked down engine wrapper, not the real thing. If you need something the API doesn't have, just ask and I'll add it, don't work around the boundary.

GTD's data is generic now too. No more `gtd_*` columns. Per account config lives in `plugin_account_config`, per message stuff in `messages.plugin_annotations`, key/value and blobs in `plugin_data`.

On the frontend, GTD registers into a slot system (right sidebar, context menu items, the row "done" button, settings, a background runtime, websocket handlers). None of the core files reference GTD anymore. One exception: the inbox tab strip (Inbox/Todo/Waiting/etc) is still in `MessageList.jsx` because it's tangled up with the category tabs, so that part still touches core.

What this means for you:
- Your old PRs were against the inline GTD code that's gone now, so they'll need to be redone against `plugins/gtd/`.
- Backend changes inside `plugins/gtd/` are safe by design, so I can merge on green CI instead of a deep review.
- GTD is opt in per user now (Settings > Plugins), off by default.

Full writeup with the whole capability list and how to test is here:
https://github.com/maathimself/mailflow/blob/main/backend/src/plugins/README.md

Happy to walk you through the API if it helps. Thanks!

PR #370, opened 2026-08-13
feat: add safe GTD classification undo

salmonumbrella:
## Summary

Ports the GTD classification undo from closed PR #333 onto the plugin boundary described in #368. Classification still copies the message into the configured GTD folder immediately; when the IMAP server returns an exact UIDPLUS destination, that new copy can now be undone without touching Inbox, another label, or another message in the thread.

## Changes

- Return whether GTD classification created a copy and, when available, its exact destination UID.
- Add a plugin-owned undo endpoint that rechecks the current state-folder mapping and matches account, folder, UID, and RFC Message-ID before deleting anything. Replays and stale tokens are safe no-ops.
- Do not offer undo when the server lacks UIDPLUS, the label already existed, or the source has no stable Message-ID.
- Add undo notifications to the GTD runtime and context menu, plus `Ctrl+Z` / `Cmd+Z` for the newest GTD classification.
- Translate the shortcut and failure state in all seven locales.
- Keep explicit unclassification unchanged. This does not port delayed archive or any other core Inbox triage behavior from #333, and it adds no migration.

## Testing

- Backend: `npm test` — 1,117 tests passed across 61 files.
- Backend: `npm run lint` and `npm run lint:plugins`.
- Frontend: full `node --test` suite, `npm run lint`, and `npm run build`.
- Regression coverage includes UIDPLUS and non-UIDPLUS copies, existing labels, missing Message-ID, exact undo, replay, stale folder mappings, forged UIDs, one-shot frontend callbacks, and the GTD undo shortcut.
- Roborev reviewed the final tree with no issues found.

Refs #368

---


maathimself:

maathimself:
Merged into main as 21b87c5 (squash), with your commit authorship preserved. I rebased onto current main and added the two missing locales (pl, cs) so the i18n parity check stays green across all 9 languages. Verified locally before merge: backend 1203 tests, frontend 1671 tests (including i18n parity), backend + plugin lint, frontend lint, and the Vite build all pass. Thanks for the careful work on the undo guards (UIDPLUS-only, Message-ID binding, state-folder re-resolution) - the destructive path held up well under review.

PR #343, opened 2026-08-01
feat: add delegated contact API

salmonumbrella:
## Summary

Adds thread-stable delegated-contact persistence and a bulk backend API, including ownership validation, immutable contact snapshots, lifecycle cleanup, read-surface enrichment, and IMAP/database compensation.

This is PR 5 of the cumulative command-system roadmap. Please review the last commit only: `2434925`.

Closes #342.

## Changes

- add migration `0040_gtd_delegations.sql` with user/account/thread ownership and contact snapshots
- add `POST /api/gtd/delegations` with deduplicated 1–100 message batches and structured partial results
- serialize per-thread COPY operations, await shared folder syncs, and compensate only exact request-owned UIDs
- reconcile metadata after manual, automatic, and external Delegated-label removal without deleting surviving assignments
- expose delegation metadata through GTD sections, message lists, single-message/thread resolution, and search
- preserve snapshots after contact deletion and reset delegation age only when the selected contact changes

## Testing

- `npm run lint` (backend, zero warnings)
- Node 22: 51 files / 1,014 backend tests passed
- focused final review suites: 212 tests passed
- isolated Compose candidate build and health check
- fresh PostgreSQL migration schema verification
- authenticated endpoint smoke tests for validation and sanitized `not_found` bulk results

---

### Contributor License Agreement

By submitting this pull request I confirm that:

- [x] I have read and agree to the [Contributor License Agreement](../CLA.md).
- [x] My contribution is my own original work (or I have identified any
      third-party material and confirmed it is compatible with the CLA).
- [x] I have the right to submit this contribution under the terms of the CLA.

PR #371, opened 2026-08-13
feat(gtd): show GTD state indicators in inbox rows

salmonumbrella:
## Summary

Ports the inbox GTD state indicators from closed PR #330 onto the plugin boundary described in #368. Ordinary inbox and search rows can now show every GTD state on the thread, with the oldest relevant classification date, without putting GTD-specific joins or fields into MailFlow's core message query.

The prerequisite plugin-boundary work from #370 is merged, and this branch is rebased onto current `main`.

## Changes

- Add a bounded, plugin-owned metadata endpoint for up to 100 visible messages at a time.
- Expose only a neutral `message-row-meta` slot from the core message list. GTD owns the labels, colors, ordering, dates, and account gating.
- Resolve state metadata across every live copy of the same message and thread, including custom GTD folder mappings.
- Group and chunk visible-row requests per account, ignore superseded responses, and keep optimistic classification and undo state coherent with server refreshes.
- Refresh indicators after GTD WebSocket events, reconnects, folder-map changes, classification, undo, and unclassification.
- Show compact state chips and a conservative age for multi-state threads.
- Treat partial metadata responses as retryable and keep retries bounded and cancelable when no consumers remain.

Refs #368

---

### Contributor License Agreement

By submitting this pull request I confirm that:

- [x] I have read and agree to the [Contributor License Agreement](../CLA.md).
- [x] My contribution is my own original work (or I have identified any
      third-party material and confirmed it is compatible with the CLA).
- [x] I have the right to submit this contribution under the terms of the CLA.

PR #372, opened 2026-08-13
feat(gtd): add delegated contact API

salmonumbrella:
## Summary

Ports the backend half of delegated-contact GTD state from closed PR #343 onto the plugin boundary described in #368. A caller can now label one or more owned threads as Delegated and attach a durable contact snapshot without adding a GTD table, migration, or command-engine dependency.

This PR is stacked on #371 and includes that predecessor commit until it lands.

## Changes

- Add `POST /api/gtd/delegations` for 1–100 unique, owned message IDs and an owned contact or `null`.
- Store contact details and a durable delegation boundary under `messages.plugin_annotations.gtd.delegation` on every live row in the thread.
- Serialize delegation, transition, and cleanup work per account and thread so concurrent requests cannot create duplicate copies or undo a successful explicit delegation.
- Reconcile non-UIDPLUS copies through one fresh destination sync and a strict UID set difference. Ambiguous outcomes are reported without guessing.
- Compensate annotation failures only when this request can identify the exact label copy it created. Pre-existing labels are never removed.
- Inherit delegation snapshots onto newly synced rows and clear them only after the final Delegated copy disappears.
- Keep personless delegation public metadata unchanged while retaining the internal boundary needed to distinguish an older message from a newer reply.

Blocked by #371.

Refs #343 and #368

---

### DIFF
--- a/backend/eslint.plugins-boundary.js
+++ b/backend/eslint.plugins-boundary.js
@@ -15,7 +15,7 @@ import globals from 'globals';
 export default [
   {
     files: ['src/plugins/*/**/*.js'],
-    ignores: ['**/*.test.js'],
+    ignores: ['**/*.test.js', 'src/plugins/gtd/**/*.js'],
     languageOptions: {
       ecmaVersion: 2022,
       sourceType: 'module',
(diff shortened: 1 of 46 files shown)


maathimself:
Queued behind #371 on #483's list. Closing until its turn.

PR #181, opened 2026-06-30
Antispam — v0.1 (Manual feedback & infrastructure)

rafpigna:
## Antispam — v0.1 (Manual feedback & infrastructure)

This is part 1 of a 3-release plan to add antispam to MailFlow: a combination of user-driven training and an optional SpamAssassin backend, both togglable per-account.

| Release | Scope | Status |
|---------|-------|--------|
| v0.1 (this PR) | User feedback & infrastructure | ✅ Ready for review |
| v0.2 | ML classifier trained on v0.1's data | Planned |
| v0.3 | Optional SpamAssassin integration | Planned |

The three releases are designed to be small and mergeable in any order without breaking each other.

The inspiration is Mozilla Thunderbird: when you connect Thunderbird to an IMAP account that has no built-in spam filter, it still *moves spam into Junk automatically* the moment a message syncs. 
This PR is part 1 of a 3-release plan to bring that experience to MailFlow.

### Why v0.1 on its own looks "useless"

At first glance this PR only exposes two new buttons — "Mark as Spam" / "Mark as Not Spam" — that move a message to / from the Junk folder. The existing "Move to folder" UI can already achieve the *visible* result, so a maintainer might wonder why the buttons exist.

The critical difference is the explicit intent: a click on "Mark as Spam" is a declaration that "this is a training sample". The backend records it in a new `spam_training_log` table and stamps the message with a verdict + override flag. That data feeds v0.2.

Without v0.1 shipping first, v0.2's classifier would start cold: it would need hundreds of labeled examples to become useful, and we'd have to wait for users to organically accumulate enough. Releasing v0.1 ahead means every user who installs it starts producing training data immediately, by doing what they'd do anyway.

#### Backend

- Migration `0019_spam_training.sql`:
  - 6 new columns on `messages`: `spam_score_sa`, `spam_score_ml`, `spam_verdict`, `spam_analyzed_at`, `spam_details`,
    `spam_user_override`
  - new `spam_training_log` table with FKs, indices, and a CHECK constraint restricting `label ∈ {spam, ham}` and `source ∈ {manual, auto_sa, auto_ml, auto_combined}`
- helper `resolveSpamFolder()` in `utils/mailUtils.js`:
  - mirrors the existing `resolveTrashFolder()` chain: `folder_mappings.spam` → special-use `\Junk` → multilingual name heuristic (spam, junk, indesiderata, spamverdacht, courrier indésirable, …)
  - 7 new tests in `mailUtils.test.js`
[...]

Issue #156, opened 2026-06-13
[FEATURE-REQUEST] Anti-Spam Engine


rafpigna:
**Is your feature request related to a problem? Please describe.**
Not a problem, but I was thinking on how would be difficult to add an antispam-engine like "Spam-Assasin" like in Mozilla Thunderbird or, at least, a "learning spam engine" that starts with a default spam rules that moves the spam in the spam folder and learns better based on the user input when the users moves a message in the spam folder or use a "not spam" button when there is a false-positive.

**Describe the solution you'd like**
I think Mozilla Thunderbird is a perfect use case and solution example. 

I know this is probably not an easy task, but I think is worth thinking about it.

maathimself:
PR #181 (antispam v0.1) has been merged and implements the first phase of this — user-driven spam/ham feedback with a training log for a future ML classifier. SpamAssassin integration is planned for v0.3.

rafpigna:
Hi @maathimself 

I'm a relatively new open-source contributor, and v0.1 was my first significant PR. I learned a lot from your review (especially the structured way you call out blocking / non-blocking / worth-noting items was really helpful). I'm trying to bring the same level of rigor to v0.2 design.

Following up on PR #181 (v0.1 manual feedback) which you merged on 2026-07-02, I've been thinking about the v0.2 architecture (ML classifier) and I want to share the design with you **before writing any code**.

The reason I'm doing this rather than jumping into implementation: v0.2 involves a number of architectural decisions that, in my view, go beyond pure technical choices, some of them touch on the *philosophy* of the project (how MailFlow should treat user data, where the line is between user control and automation, how configurable things should be). I want to make sure that the v0.2 I build aligns with your vision for MailFlow, not just my interpretation of it.

I wrote down an ADR-like (Architecture Decision Record) that covers the main decisions I'd make for v0.2, with the rationales and the alternatives I considered for each. It's intentionally a discussion document, not a finished design. I'm fully open to changing anything you disagree with, adding things I may have missed, or cutting things you think are over-engineered.

I've also added **7 specific questions for you** in §10. These are the points where I think your input matters most, and where my personal preference might not align with what you'd want for MailFlow.

What I'd love from you at this stage:

1. **A high-level sign-off** (or push-back) on the overall approach
2. **Answers to the 7 open questions** (or just "I don't have a strong opinion on this, your call")
3. **Any additional considerations** I may have missed
[...]

Issue #130, opened 2026-06-01
Drag Message to Folder (or Folder Favorite) to Move Message (Desktop)


computrav:
**Is your feature request related to a problem? Please describe.**
I was using the app this morning on a desktop browser and instinctively went to drag a message into folder on the left pane, and realized it's not possible to drag messages. :)

**Describe the solution you'd like**
- Click and drag messages into a folder or into a favorited folder.
- If dragging onto a collaposed folder tree, auto-expand the tree after X seconds of hover
- Complete the move when the user on click-up

maathimself:
Implemented in dab4d78 (draggable/resizable compose) and 98fbb7f (drag to folder).

computrav:
I think as implemented, you only intended one message to be moved at a time, but I wasn't sure, because when you move the message and see the status message (with undo...nice) at the bottom, it says "1 message moved". This lead me to test moving more than 1 message to confirm if the status message updates as expected. By doing this I realized that it's not possible to move multiple messages (if you've selected more than one). 

So I think one of these two things may want to be considered:
1. Restrict moving to one message and then update the status message accordingly (from "1 message moved" to "Message moved")
2. Allow moving of one or more messages and update the status message ("1 message moved" / "X messages moved" if > 1)

I think #2 is the most expected UI behavior.

maathimself:
Multi-message drag implemented in 72f1a9f — all selected messages move together with correct count in the undo toast.

braadaaay:
Hi @maathimself,

I'm having trouble with drag and drop (both single and multiple). I'm using Microsoft Edge 152.0.4191.66. Do I need to enable a setting somewhere, or is it possible there has been a regression?

Cheers!

maathimself:
Hi @braadaaay, nothing in the drag-and-drop code has changed recently, so a regression looks unlikely,
but there's a known rough edge that would produce exactly what you're describing.

Dragging is disabled whenever the browser window is narrower than 768px wide. That's a
viewport-width check meant to switch to the mobile layout, and it doesn't distinguish a
phone from a desktop window that simply isn't very wide. If your Edge window is snapped
to half a screen, or the zoom level is high enough to bring the effective width under
768px, drag will silently do nothing for both single and multiple selections.

Could you check whether it works with the window maximized at 100% zoom? That would confirm
or rule it out quickly.

If it still fails maximized, a few things would help narrow it down:
- Does the message row show a drag preview when you start dragging, or does nothing happen at all?
[...]

PR #388, opened 2026-08-19
feat: add ML-based antispam classifier (v0.2)

rafpigna:
## Summary

This PR implements the **v0.2 ML classifier** for the antispam filter, following the plan discussed in #156 (v0.1 manual feedback → v0.2 Bayesian ML → v0.3 SpamAssassin). It builds on the `spam_training_log` infrastructure shipped in v0.1 (PR #181): when a user marks a message as spam/ham, the message features are now captured **at mark time** (Solution C — no JOIN on the messages table) and a per-user **Multinomial Naive Bayes** model is trained and used to auto-classify incoming messages into the Junk folder.

The classifier is deliberately a **hybrid 3-layer** system, designed so that v0.3 (SpamAssassin) can slot into the same interface later:
1. **Manual user override** — always wins, never overridden by the model.
2. **Per-user MNB model** — active once a user has ≥ 50 training records.
3. **14-rule heuristic engine** — always on, and the only scorer below 50 records.

Because this is a self-hosted product with the user's own mail data, the design puts **privacy and auditability first** (see the threat model in Testing).

## Review round (2026-09-13)

Since the first review I rebased onto `main` (the branch is now based on the current `main` tip) and addressed all three points from the review, in this PR:

- **Authserv-id trust gate** — `Authentication-Results` is now honored **only** when the header's authserv-id matches `email_accounts.trusted_authserv_id` (new column, migration `0056`). With no value configured (the default) the header is treated as absent: no pass weights, no `AUTH_*_FAIL` silencing. The authserv-id is parsed per header instance — including the shape the ingest parser produces, where repeated headers are joined with `\n` under one key — so a forged header can no longer shadow the real MTA's result. Because typing an authserv-id is awkward for self-hosted users, the account form shows the ids **actually observed** on that account's recently classified mail (with counts) and applies one with a click (`GET /api/spam/authserv-ids`); nothing is trusted automatically, since an id that appears on a minority of messages is more likely forged than genuine.
- **Size cap before tokenizing** — `tokenize()` now caps each raw field at **64,000 characters** (applied before the HTML parse, which is the expensive step) and the token stream at **2,000 tokens**; subject tokens are pushed first so they always survive. The same caps apply on the training path, so the features stored at mark time match what inference sees.
- **Non-UIDPLUS destination guard** — the auto-move path now mirrors the manual `/spam` path: when the server returns no new UID, the destination UID is guarded and released after the same 10 s grace period, so an auto-moved row can no longer be dropped by the next reconcile.

The E2E round also surfaced two defects in my own code, both fixed here: the ambiguous verdict band was persisted as `uncertain`, which violated the existing `messages.spam_verdict` CHECK (spam|ham|unsure|pending) and silently dropped the verdict — it now persists `unsure`, the vocabulary the schema already defines; and the ingest hook lived only in `syncMessages`, so messages inserted by the **backfill** path (initial population, manual reindex, UIDVALIDITY recovery) were never classified at all — the hook is now shared by both paths.

## Review round 2 (2026-09-15)

Rebased onto the current `main` (28 commits on top of `11c55c0` / v3.4.2) and addressed both points from the second review:

- **No auto-move on the backfill path** — `classifyAndTagMessage` takes `deferAutoMove`, set only by the backfill call site. The message is classified and tagged (verdict, score and `spam_details.autoMoveDeferred` all persisted), but never moved: a reindex classifies a whole mailbox at once, and every would-be move went through the 2-slot connection pool, each overflow loser opening a fresh login at the provider. New mail is still auto-moved on the sync path. The reindexed backlog therefore stays in place, tagged, and the persisted `autoMoveDeferred` flag keeps a deliberate, user-triggered catch-up pass cheap to add.
- **Auto-moves serialized per account** — one move in flight per account, the rest queued FIFO behind a bounded wait (`createKeyedSemaphore`, the primitive suggested in the review). Classification stays concurrent; only the move is queued. This also covers the folder-status-driven ingests, which classify up to 100 messages in one pass.
- **Retrain scheduler re-entrancy** — one run at a time across the hourly bucket and the admin retrain, with an overlapping run refused rather than queued (409 on `POST /api/spam/retrain-now`, mirroring the inbox-rules sweep guard). The hourly tick is a self-scheduling timeout armed only after the current run settles, so a slow bucket delays the next tick instead of overlapping it, and each user's retrain is bounded by a timeout so one user cannot monopolize a run.
- **Frontend conflicts** resolved by union (three files, four hunks). One of them was not a plain import: `ThreadRow`'s signature conflicted on `onDragStart` vs `onExplainSpam` while the body already used both — taking either side alone would have regressed the other.
- **`spam_training_log` retention** is deliberately out of scope here and will land as a follow-up: the bodies are stripped after a configurable window (default 90 days, per-user, 7–365) while the token counts and flag features are kept, so the training signal survives and `training_records` does not drop below the 50-record line.

### Backend
- **Schema** (migrations 0052–0056, appended after the upstream 0049–0051 with no gap):
  - `0052_spam_training_features` — captures subject/body/token counts/flags/sender/attachments at mark time (Solution C).
  - `0053_spam_models` — per-user model state: JSONB vocabulary, weighted totals (`REAL`, since the decay retrain produces fractional counts), priors, decay window, model version.
  - `0054_email_accounts_antispam_enabled` — per-account toggle (default OFF).
  - `0055_spam_training_deletions` — GDPR audit trail for training-data resets.
  - `0056_email_accounts_trusted_authserv_id` — the authserv-id whose `Authentication-Results` headers are trusted (NULL = trust none).
- **Core ML** (`src/services/spam*.js`):
  - `spamTokenizer` — bag-of-words (subject ×2, HTML strip, unicode CJK/Cyrillic, URL-host tokens, stop-words in 7 languages, heuristic flag features, message fingerprint) + the size caps described above.
  - `spamParser` — `Authentication-Results` parsing per RFC 7601/8601, with the authserv-id trust gate (`extractAuthservId`, `hasTrustedAuthResults`, and an `extractAuthservIds` helper used for the observed-id report).
  - `spamRules` — the 14 heuristic rules approved in the design doc (attachment dedup uses max, not sum). The auth rules are neutral unless a **trusted** header is present.
  - `spamModel` — MNB scoring in log-space: priors derived from the training set, **fixed asymmetric auth weights** (dkim/spf fail +0.7, dmarc fail +0.5, pass −0.2/−0.1/−0.1), chi-square feature selection (cap 10k vocab), confidence blending.
  - `spamModelStore` — DB I/O with a 5-minute in-process cache; `spamScheduler` — hourly retrain with per-user staggered bucket (`hash(user_id) % 24`) and configurable exponential decay (default 90 days, range 7–365).
  - `spamPipeline` — classifies new messages on ingest (shared hook in `imapManager`, used by both the sync and the backfill path), auto-moves to Junk at ≥ 0.95 **only when the ML is active**, persists verdict/spam_details (including the observed/trusted authserv-ids). Auto-verdicts are never written back to the training log (prevents self-poisoning; only manual feedback trains).
  - `spamAuthservIds` — reads the ids recorded in `spam_details` back out for the account-form helper.
- **API** (`src/routes/spam.js`): status, per-user thresholds (GET/PATCH, admin), decay threshold (7–365), admin retrain-now, master enable, GDPR reset (per-account + global, with explicit confirm), deletions audit, a `Why?` explain endpoint (fired rules + top ML tokens), and `GET /authserv-ids` (admin) for the trusted-id setup helper. Two-level enable: per-user master switch + per-account toggle; the pipeline does a `JOIN users` and skips when the master is off. `trusted_authserv_id` is validated/normalized on the account update route.

### Frontend
- New **Antispam tab** in the Admin panel: model status, score-slider thresholds, decay control, master switch, admin retrain, GDPR reset with confirm dialog.
- **SpamExplainModal** ("Why?") — shows which rules fired and the top ML tokens behind a verdict.
- **SpamBadge** — clickable `spam 98%` badge on message rows/panes (shown only for *auto* verdicts, so a manual override never looks like a model decision).
- Per-account `antispam_enabled` toggle plus the **trusted authserv-id field with the observed-value helper** in the account form.
- i18n strings for all **9 locales** (en/it/de/es/fr/ru/zhCN + cs/pl).

## Testing

- **Unit**: backend **1586/1586** (89 files), frontend **2007/2007** (191 files), lint clean on both, plugin-boundary lint clean, frontend build clean. New tests cover the trust gate (forged all-pass header ignored, trusted fail honored, case-insensitivity, both header shapes), the size caps (including a truncated `<script>` that must not leak), the auto-move guard on both branches, the verdict vocabulary, and the ingest hook.
- **E2E (greenmail, isolated stack)** — I ran the pipeline against a throwaway copy of the stack, so nothing touched the production instance. **12/12 scenarios**:
  - fresh install (rules-only, no auto-move) ✅
  - manual spam mark → override persists, re-analysis skips the message ✅
  - ≥50 training records → blended score ≥ 0.95 → auto-move to Junk on the **sync** path (the real ingest path, `POST /mail/sync`; the earlier round exercised it through a reindex) ✅
  - reindex (backfill) with the model active → tagged `spam` but **not** moved, `autoMoveDeferred` persisted ✅
  - scheduler single-flight: two concurrent full retrains → one runs, one refused ✅
  - borderline ham stays in place ✅
  - **authserv-id gate**: with no trusted id the forged all-pass header is ignored (no `AUTH_*` rule fires, `authTrusted=false`); with `trusted_authserv_id = mx.example.com` the real `dkim=fail`/`spf=fail` win over the forged `pass` ✅
  - **observed-id helper**: `GET /api/spam/authserv-ids` reports both ids seen with counts ✅
  - **size cap**: a ~5 MB message is classified in ~8 s with the backend still healthy ✅
- **Migration upgrade path** — bootstrapped a clean Postgres with the upstream migrations `0001`–`0051` plus a realistic v0.1 dataset (user, account, message, `spam_training_log` row), then ran this branch's migration runner on top: it applied **only** `0052`–`0056`, `schema_migrations` ended at 56, the new columns/tables exist, the pre-existing row survived untouched, and the backend boots with the spam routes live. Fresh installs and upgrades were both exercised.
- **Performance**: tokenize + flags 0.17 ms/msg on a typical 4 KB message and **1.9 ms/msg on a 5.4 MB message** (the caps bound the work); classify ~2 ms/msg at 10–20k vocab.
- **Manual UI walkthrough** (desktop + responsive/mobile): antispam tab, thresholds/decay persistence, GDPR confirm, badge in list + pane + thread, "Why?" modal, per-account toggle, trusted-id field with the observed-value chips.
- **Security / threat model** for `/api/spam/*`: none of the endpoints are public — 5 layers: (1) global CSRF guard on `/api` (`X-Requested-With`, 403 without it); (2) `requireAuth` on every spam router; (3) `requireAdmin` on threshold PATCH, retrain-now and the authserv-id report; (4) ownership check on explain, per-account reset and the authserv-id report; (5) explicit `confirm:true` on GDPR resets. The trusted-authserv-id value is length- and charset-validated before it reaches the database. Residual, intentional: a non-admin may read their own status and delete their own training data (a GDPR right). Future hardening: per-user rate-limiting on resets/retrain-now; IP address is already audited in `0055`.
- **Migration numbers** were re-validated against `main` immediately before each push (upstream max is still `0051`; the branch's set starts at `0052`, contiguous with no gap).

---

## Design decisions: antispam toggle semantics (edge cases)

This section pins down what enabling/disabling the antispam feature does, so the
behavior in these edge cases is explicit and reviewed rather than discovered.

**Two-level enablement.** There are two independent switches, both gating
automatic classification: a per-user **master switch**
(`users.preferences.spamEnabled`, default on) and a **per-account toggle**
(`email_accounts.antispam_enabled`, default off / opt-in). The pipeline returns
`skipped` before scoring if either is off.

**A toggle controls the whole automatic pipeline, not just auto-move.** Turning
antispam off disables automatic scoring, auto-move to the spam folder, *and* the
auto-spam badge on new messages. It does **not** disable manual spam/ham marks —
those still train the model on any account.

**The badge is historical, not a live toggle mirror.** `SpamBadge` renders from
the persisted `messages.spam_verdict` written at analysis time. So:

| Situation | Already-present messages | New messages after the change |
|-----------|--------------------------|-------------------------------|
| Per-account toggle turned OFF | Keep badges on ones already tagged spam (history). No re-analysis. | Not analyzed → no badge, no auto-move. Other accounts unaffected. |
| Master switch turned OFF | Keep badges on already-tagged spam (history). | Not analyzed → no badge, no auto-move in any account. |
| Turned back ON | Keep existing verdict/history; no backfill. | Analyzed from now on → new badges/auto-move as confidence dictates. |
| Manual spam/ham mark | Sets `spam_user_override`; badge hidden for that message. | Same; manual marks always feed training regardless of toggles. |

I deliberately chose the historical behavior: it does not "re-legitimize" months
of auto-filtered mail because a toggle was briefly flipped, does not erase
history on disable, and does not retroactively "pop" badges on re-enable
(`spam_verdict` is a property of the message, like `category`). If you'd prefer
the badge to mirror the live toggle state instead, that's a small isolated
frontend change — happy to switch on request.

---

## Note: screenshots and user documentation

The UI walkthrough screenshots are shared as a public zip rather than embedded
inline, to keep the diff focused on code:

**📦 [antispam-v0.2-screenshots.zip](https://drive.raffaelepignataro.it/index.php/s/sZfwY4E8HYor4rL)**

They cover the Antispam tab, thresholds/decay, the per-account toggle, the
auto-spam badge (list + pane), the "Why?" explain modal, and the GDPR reset
confirm dialog.

On documentation: I noticed this repository currently has **no `docs/` directory
or end-user/help documentation** — the only markdown beyond the code is
`README.md`, `CONTRIBUTING.md`, `ROADMAP.md` and `CLA.md`, even though the
official site does ship user docs at https://mailflow.sh/docs. Rather than
introducing the repository's first end-user docs as part of this (already large)
feature PR, I kept the antispam "user guide" out of scope. I'd welcome a quick
steer on whether you'd like the antispam UI documented in the docs site and/or
as a new e.g. `docs/antispam.md` here — happy to prepare that in a small
separate PR if you confirm the preferred location/format.

---

---

## On AI assistance

AI tools and LLMs were used to speed up development, scaffolding, code generation,
bug fixing and analysis. All key decisions about architecture, design and project
strategy, as well as the real-world E2E tests on actual usage scenarios, were made
and supervised by a human.

---

— r.


maathimself:
Thanks for putting this together, and sorry for the delay. I am prioritizing reliability and stability over new features right now, so I am not going to merge new-feature work at the moment. I am keeping it open for when I return to feature work, but please expect that to be a while.

One extra note: because this touches security-sensitive areas, it will also need a careful security review before it could be merged, so it is not only a scheduling delay. Appreciate the effort you put in.

rafpigna:
> Thanks for putting this together, and sorry for the delay. I am prioritizing reliability and stability over new features right now, so I am not going to merge new-feature work at the moment. I am keeping it open for when I return to feature work, but please expect that to be a while.
> 
> One extra note: because this touches security-sensitive areas, it will also need a careful security review before it could be merged, so it is not only a scheduling delay. Appreciate the effort you put in.

No problem at all.
Let me know when you will be ready to evaluate the merging and how I can help with the security review when you think it will be the right time.

maathimself:
I did a full security + reliability pass. The auth model checks out (CSRF, requireAuth everywhere, requireAdmin re-verified against the DB, no IDOR, parameterized queries throughout), no XSS in the "Why?" UI, no catastrophic regexes, the anti-self-poisoning design holds, the migrations are idempotent and don't collide, and the ingest hook is properly isolated so a classify failure can't break sync. Nice work, and it fails safe (opt-in, default off, auto-move is recoverable).

Two things I do need before I can merge:
- The CLA checkboxes in the PR description are unchecked, I need those confirmed.
- The branch is behind main now (main is at migration 0050), so it needs a rebase, and please renumber the migrations to close the resulting gap.

And three things I'd like addressed, either in this PR or as quick follow-ups, your preference:
1. `Authentication-Results` is trusted without checking its source. Because repeated headers get merged and `pass` wins across them, a sender can add their own `Authentication-Results: x; dkim=pass; spf=pass; dmarc=pass` and it overrides the real MTA's `dkim=fail`, earning the pass weights and silencing the AUTH_*_FAIL rules. Worth only honoring results whose authserv-id matches a trusted per-account value, and treating them as neutral otherwise. (I know the trusted-authserv-id default is awkward for self-hosted users, happy to talk through it.)
2. The auto-classify path has no body/token-size cap before tokenizing. It's fire-and-forget so it won't block sync, but a size cap would keep a huge message from doing unbounded synchronous work.
3. `autoMove` doesn't carry the destination-UID guard that the manual /spam path uses for non-UIDPLUS servers, so on those servers an auto-moved verdict can be lost on the next reconcile. Mirroring the manual path's guard closes it.

None of these block the concept and the design is sound. Let's get the CLA and rebase sorted and take it from there. Thank you.

rafpigna:
Sorry for late reply, I just noticed it was in review just now. I will sort the cla, rebase and the open points ASAP.

rafpigna:
Hello @maathimself 
Thanks — the CLA, the rebase and all three points are sorted in the branch.

**CLA**: the three boxes in the PR description are checked.

**Rebase**: rebased onto the current `main`. The migrations are `0052`–`0056` and now follow the upstream `0049`–`0051` with no gap, so the renumbering you asked for is in place. I also picked up the two locales that landed upstream meanwhile (cs, pl): the antispam strings now cover 9 languages.

**1. `Authentication-Results` trusted without checking its source** — fixed as you suggested. The header is only honored when its authserv-id matches a trusted per-account value (`email_accounts.trusted_authserv_id`, new column in `0056`). With nothing configured — the default — the header is treated exactly as absent: no pass weights, and no `AUTH_*_FAIL` silencing. The id is extracted per header instance, including the shape my ingest parser produces (repeated headers joined with a newline under one key), which is exactly where a forged header could shadow the real MTA's result. I also had to make the rules' "is an Authentication-Results header present?" test trust-aware, not only the parsed values: the negative-result sets include `null`, so an untrusted header would otherwise still have fired the AUTH rules.

On the awkward default you flagged: I kept NULL = trust nothing (fail safe) and added a helper in the account form that lists the authserv-ids actually observed on that account's recent mail, with how often each appears, and applies one with a click. It is advisory only — nothing is trusted automatically, because an id that appears on a minority of messages is more likely forged than genuine. Happy to change the default if you prefer a different trade-off.

**2. No size cap before tokenizing** — the tokenizer now caps each raw field at 64,000 characters (applied before the HTML parse, which is the expensive step) and the token stream at 2,000 tokens; subject tokens are pushed first, so they always survive the cap. The same caps apply on the training path, so the features stored at mark time match what inference sees. A 5.4 MB message now costs ~1.9 ms instead of unbounded synchronous work.

**3. `autoMove` destination-UID guard** — mirrored from the manual path: when the server returns no new UID, the destination UID is guarded and released after the same 10 s grace period, so an auto-moved verdict survives reconcile on non-UIDPLUS servers.

**Two bugs the E2E round caught in my own code**, fixed here as well:

- the ambiguous verdict band was persisted as `uncertain`, which violates the existing `messages.spam_verdict` CHECK (`spam|ham|unsure|pending`) — the UPDATE was rejected and the verdict silently dropped. It now persists `unsure`, the vocabulary the schema already defines.
- the classification hook only lived in the sync path, so messages inserted by the **backfill** path (initial population, manual reindex, UIDVALIDITY recovery) were never classified at all. The hook is now shared by both paths.

**Testing**: backend 1533/1533, frontend 1926/1926, lint and build clean. E2E on an isolated copy of the stack: fresh install (rules-only, no move), manual mark + override, ≥50 records → auto-move to Junk, borderline ham untouched, the authserv-id gate (with a forged all-pass header the trusted `dkim=fail`/`spf=fail` still win; with no trusted id the forged header is ignored), the observed-id endpoint, and a ~5 MB message classified with the backend healthy — all green. Plus an upgrade-path run from upstream `0001`–`0051` with a v0.1 dataset on top: only `0052`–`0056` applied, `schema_migrations` reached 56, the pre-existing row was untouched and the backend booted with the spam routes live. The auto-move guard is unit-tested only — greenmail speaks UIDPLUS, so I could not exercise the non-UIDPLUS branch end to end.


If there is something else to sort out or to change, please let me know.

maathimself:
Thanks for working through all of that, and for the detail in your write-up.

I re-checked the three points rather than just taking the summary, and they're properly
done. The authserv-id gate is right, including making the rules' "is a header present"
test trust-aware — that's the part that's easy to miss. The tokenizer caps
(MAX_BODY_CHARS 64,000 / MAX_TOKENS 2,000, applied before the HTML parse) are in the
right place, and the training path uses them too so stored features match inference. And
the destination-UID guard is mirrored with a matching unguard. CLA and migration
numbering are sorted.

Two things left, and to be clear about one of them: it's in code that arrived after my
August review, so this isn't me moving the goalposts.

**1. Auto-move during backfill.** The new backfill hook means a mailbox that just enabled
antispam gets its whole history classified on reindex. `autoMove` goes through
`moveMessage` → `withFreshClient` → `acquirePooledClient`, and POOL_SIZE is 2, so those
moves compete for the same two connections the backfill's own fetches need, with a 10s
overflow timeout for the losers. On connection-sensitive providers that's the exact
pattern I've spent the last few weeks getting out of.

Simplest answer is to tag but not move on the backfill path, and let a later sync do it.
If you'd rather keep the move, `createKeyedSemaphore` in imapManager already exists for
bounding fan-out per account.

**2. Scheduler re-entrancy.** `setInterval` fires `runBucket` hourly and it awaits
retrains sequentially with no timeout, so a long bucket overlaps the next one.
`runFullRetrain` behind the admin button has no guard either, so it can be double-clicked
and can overlap the hourly job. `folderStatus` has a `running` map that solves this.

One note, no action needed now: `spam_training_log` keeps full body_text and body_html per
marked message with no retention policy. Growth is bounded by user action rather than mail
volume so it's slow, but it's worth a deliberate decision rather than defaulting to
forever.

On the conflicts — you did rebase, and it's since drifted because of what landed on main
today, including two of the three conflicting files. That's on my end, not yours. Happy to
hold main steady for a day if you tell me when you're rebasing.

rafpigna:
@maathimself  I'm really sorry for still not catching all the edges, this project had grown a lot and because my limited experience in so big projects and mailflow complexity, make me miss some edges so thanks a lot for your detailed review each time. 

Absolutely agree on both points and I will clear that as you suggested. 
About the `spam_training_log` you are right, this was in my to-do items when I started design of the v0.2 but this topic got unaddressed in the last revisions I made, so I will investigate how to solve this possible unbounded growth without adding complexity for the end user maybe in a followup and not in this PR scope to avoid adding new items.

I will work on all the points just now (September 15 @ 20:00 UTC) so expect an update in the next few hours not later than September 16th @ 02:00 UTC. As I always do, since mailflow is so active, I will do a rebase just a minute before the push, so if nothing important changes in the next few hours, should be no need to hold main. If after my push we still see conflicting caused from what dropped in main since my push, we will agree on a small time frame when I can drop a new rebase that can be merged clean.

rafpigna:
Both points are fixed, and both were real — the backfill hook and the scheduler are mine, so
neither is a goalpost move.

**1. Auto-move during backfill.** I took your simplest answer: the backfill tags, it never moves.
`classifyAndTagMessage` now takes `deferAutoMove`, and only the backfill call site sets it. The
verdict, the score and the fact that a move was warranted all still get persisted
(`spam_details.autoMoveDeferred`), so nothing is lost — the message just stays where it is with
its badge.

Being explicit about the consequence, because it is not what "let a later sync do it" gives you
today: once the row exists, nothing re-examines it. A later sync only ingests UIDs it has not
seen, so the reindexed backlog will never be moved by a subsequent sync — it stays in place,
tagged. I record the deferred intent precisely so that a deliberate catch-up pass is cheap to add;
if you want the backlog moved, I would rather do it as a bounded, user-triggered drain
(serialized on the same semaphore as below) than as something automatic, since moving hundreds of
historical messages during a reindex is the part you objected to.

I also added your semaphore suggestion, because the fan-out is not limited to the backfill path:
auto-moves are now serialized per account (`createKeyedSemaphore(1)`, one move in flight, the rest
queued FIFO behind a bounded wait). Classification stays fully concurrent — only the IMAP move is
queued. That also covers the folder-status-driven ingests, which classify up to 100 messages in a
single pass, so those could previously fan out the same burst of moves and overflow logins.

**2. Scheduler re-entrancy.** One run at a time, held for its whole duration, covering both the
hourly bucket and the admin retrain. An overlapping run is refused rather than queued — a full
retrain already covers every user a bucket would touch — and `POST /api/spam/retrain-now` answers
409 now (I used the same shape your rules sweep uses; the disabled button only covers one tab).
The hourly tick is a self-scheduling timeout armed only after the current run settles, so a slow
bucket delays the next tick instead of overlapping it — setInterval was the overlap, and a plain
skip-if-running guard would have turned it into silently lost hours. Each user's retrain is also
bounded by a timeout (the queries are already bounded by the pool's statement_timeout, this bounds
the JS-side await too), so one pathological user cannot monopolize a run that has to finish before
the next tick.

**On the conflicts**: the three frontend files were all additive (imports, plus one component
signature), and rerere now has the resolutions recorded. Worth flagging one that is not obvious
from the markers: `ThreadRow`'s *signature* conflicted on `onDragStart` (yours, #130) vs
`onExplainSpam` (mine), while the *body* had auto-merged and already used `onDragStart` — taking my
side of the signature would have quietly made conversation rows non-draggable again. It is a union
of both props.

**On `spam_training_log` retention**: agreed it deserves a decision rather than a default, and I am
not doing it in this PR. The decision I will implement as a follow-up: `body_text`/`body_html` are
stripped after a configurable window (90 days by default, per-user, 7–365, with a keep-forever
option) while `token_counts`, `flag_features` and the sender/attachment columns are kept — so the
training signal survives and `training_records` does not drop, which matters because losing rows
would silently push a user back under the 50-record line and switch the model off. Rows from the
v0.1 era, where the raw text is the only copy of the signal, are never touched.

**Verified**: backend 1586/1586, frontend 2007/2007, lint and plugin-boundary clean, frontend build
clean. E2E against a throwaway copy of the stack: 12/12, including two scenarios I added for this round — a reindex with the model active
asserts the message stays put with `autoMoveDeferred`, and two concurrent full retrains assert one
runs and one is refused. The auto-move scenario now goes through `POST /mail/sync` instead of a
reindex, which means it exercises the real ingest path for the first time. Migration upgrade path
11/11 (no new migrations; 0001–0056, `schema_migrations` = 56). The non-UIDPLUS destination guard
is still unit-covered only: greenmail speaks UIDPLUS, so that branch cannot be reached end-to-end.

— r.

maathimself (approved):
lgtm

PR #394, opened 2026-08-22
fix(pane): stop email links opening duplicate tabs on repeated clicks

Monkey7539:
## Summary

Clicking a link in an email opens the expected new tab — but the next link click opens 2 tabs, the one after that 3, and so on, one extra duplicate per click since the message was opened.

Root cause: the iframe's link-click interceptor (which forwards `http(s)`/`mailto` links to `window.open` so relative hrefs can't resolve against the MailFlow origin) is attached as an anonymous listener inside `onLoaded` and never removed. The effect that installs it depends on two callbacks (`hasNativeContextTarget`, `openPaneContextMenu`); when their identity changes — which the tab switch caused by clicking a link reliably triggers — the effect re-runs, finds the iframe document already in `readyState === 'complete'`, and calls `onLoaded()` again on the same document, stacking one more click listener. Each stacked listener fires its own `window.open`.

## Changes

- Track the click handler in the effect scope (exactly like the neighbouring `contextmenu` handler already is) and remove it in the effect cleanup.
- Guard both the click and contextmenu handlers against double-attach when `onLoaded` re-runs against the same document.

## Testing

- Reproduced on a live self-hosted instance (source build): open an email, click several links in sequence — tab count per click grows 1, 2, 3…
- After the fix: each link click opens exactly one tab regardless of how many links were clicked before, across message switches and pane re-renders.

---


maathimself (approved):
lgtm

PR #395, opened 2026-08-22
fix(sidebar): create subfolders under their parent, not as siblings

Monkey7539:
## Summary

Creating a subfolder from the sidebar ("New subfolder" on a folder's context menu) actually creates a **root-level folder** — it appears as a sibling of the intended parent, even though the create input renders indented underneath it.

Root cause: `handleCreateFolderSubmit` calls `api.createFolder(accountId, name)` and drops `creatingFolder.parentPath` — the very state that positions the input row under the parent. The API helper already accepts a `parentPath` third argument, and the backend route already looks up the account's IMAP delimiter and builds the full `parent<delim>name` path; the value just never left the component.

## Changes

- Pass `creatingFolder.parentPath` through to `api.createFolder`. Root-level creation (where `parentPath` is undefined) is unchanged.

## Testing

- Reproduced on a live self-hosted instance (source build) against a Gmail account: "New subfolder" on any folder created a top-level folder.
- After the fix: subfolders are created as true IMAP children (verified for nested parents several levels deep and confirmed server-side via the folder list after resync); root-level "New folder" still works as before.

---


maathimself (approved):
lgtm

PR #396, opened 2026-08-22
fix(context-menu): re-clamp to viewport when the menu grows

Monkey7539:
## Summary

The message context menu can render partly off the bottom of the screen, most noticeably in the move-to-folder flow: open the menu directly in move mode (row folder icon) on a message near the bottom of the list, and the folder list extends past the viewport edge where it can't be scrolled to or clicked.

Root cause: the viewport-fit adjustment measures the menu once, when it opens (the effect depends only on `[x, y]`). The menu's height then changes after that measurement — the move view starts as a short "Loading folders…" panel that fits below the click point and grows when the folders arrive, and switching from the actions list into the Move/Snooze/Categorize subviews swaps in taller content the same way. Nothing re-clamps after the growth.

## Changes

- Observe the menu with a `ResizeObserver` and re-run the existing clamp logic on every size change, so a menu that grows near the bottom edge flips above the anchor point instead of overflowing. No change to the initial placement behaviour.

## Testing

- Reproduced on a live self-hosted instance (source build): move-to-folder on a bottom-of-list message rendered below the fold; after the fix it flips up and stays fully on-screen. Verified the normal open-then-navigate flow (right-click → Move to folder / Snooze) near the bottom edge as well, and that menus opened mid-screen are unaffected.

---


maathimself (approved):
lgtm

PR #397, opened 2026-08-22
fix(message-list): keep the bulk-move picker on-screen

Monkey7539:
## Summary

The bulk-selection toolbar's move-to-folder picker can render with its left side clipped (folder names cut off mid-word) or partly off-screen.

Root cause: the picker is `position: absolute; right: 0` inside its toolbar button, and that button sits near the left edge of the message-list panel. The popover (up to its 320px max-width) therefore extends leftward past the panel edge, where an ancestor with overflow clipping cuts it off. How visible this is depends on folder name lengths and panel width, so it surfaces intermittently.

## Changes

- Position the picker `fixed`, anchored to its toolbar button — the same approach the layout picker in this file already uses — and clamp it to the viewport on all sides (right-aligned to the button when it fits, slid right when it would poke off the left edge, flipped above when there's no room below).
- Re-place it on size changes via `ResizeObserver`, since its dimensions change as folders load and as search filtering grows/shrinks the list.
- Mobile keeps its existing full-screen sheet; only the desktop popover is touched.

## Testing

- Reproduced on a live self-hosted instance (source build): select several messages in a narrow message-list layout and open the move picker — left side clipped. After the fix the picker renders fully visible under the button, and stays on-screen while searching and while the folder list loads.

---


maathimself (approved):
lgtm

PR #398, opened 2026-08-22
fix(sidebar): keep single-child folders visually indented

Monkey7539:
## Summary

In the account sidebar, a subfolder that is its parent's **only child** renders flush with the parent — visually a sibling — even though collapsing the parent correctly hides it. Folders with two or more children indent normally, which makes the inconsistency look like a data problem to users when it's purely visual.

Root cause: the drag-to-reorder handle introduced with folder reordering only renders when a folder has `siblings.length >= 2` (nothing to reorder otherwise). An only-child row therefore loses the handle's footprint (~16px including the flex gap), which almost exactly cancels the 14px per-level depth indent.

## Changes

- Render an invisible spacer with the handle's width on rows that can't be reordered (desktop only — mobile never shows handles, so it's already consistent there). Indentation is now driven purely by depth for every row.

## Testing

- Reproduced on a live self-hosted instance (source build) with a Gmail account: any folder with exactly one subfolder showed the child un-indented; after the fix it sits one clean step in from its parent, and multi-child groups are unchanged.

---


maathimself (approved):
lgtm

PR #408, opened 2026-08-26
fix: preserve mail sync and GTD thread integrity

### LINKED ISSUE #407: Prevent transient ghost rows and unsafe GTD completion during IMAP reconciliation
## The situation

Rapid archive activity can overlap with IMAP folder reconciliation. During that window, `GET /api/mail/messages` can expose rows whose UID is known but whose envelope has not been fetched yet. The inbox renders those rows as `Unknown / (no subject)` until reconciliation catches up.

The same identity gap affects destructive operations. GTD Done can remove a state label and archive only part of a thread when folder copies change underneath the request, while retries can repeat provider work after an uncertain response.

Normal, serialized archive and GTD flows usually converge after sync. That is not enough during rapid input, reconnect-driven refreshes, provider retries, UIDVALIDITY changes, or partial backfills. Local state can temporarily outrun provider identity while the HTTP requests still succeed.

## Ask

Make message visibility and provider mutations depend on complete, verified folder and message snapshots. Persist mutation intent and receipts so retries resume the same operation. Make GTD Done finish one frozen thread-wide worklist before reporting success, preserve legacy snooze identity during migration, and defer destructive folder-wide work until the folder is authoritative.

salmonumbrella:
## What changed

- Keep incomplete IMAP envelope rows out of mail, search, category, rule, and GTD read surfaces until backfill verifies them.
- Fence moves, copies, deletes, appends, drafts, sends, flags, and forwarding with exact folder and message snapshots plus durable provider-operation receipts.
- Make rapid archive requests idempotent and reconcile uncertain subsets without replaying completed work.
- Make GTD Done freeze and resume a thread-wide plan: mark every copy read, archive Inbox copies, then remove the requested GTD labels.
- Preserve safe GTD classification undo on the durable COPY path.
- Bind legacy snoozes before topology quarantine, establish UIDVALIDITY before using a newly ensured folder, and defer empty-folder while its backfill is incomplete.
- Repair WebSocket URL construction and reconnect refresh handling so live updates do not degrade into the polling loop that exposed the race.

## Why

An IMAP UID is meaningful only inside one folder and one UIDVALIDITY epoch. During rapid archive bursts or resyncs, MailFlow could list a UID before its envelope was complete, then let later mutations act on stale local identity. That produced transient `Unknown / (no subject)` rows and made archive and GTD retries ambiguous.

This change makes incomplete metadata invisible, carries exact snapshot identity through provider work, and stores enough intent and receipt state to recover without guessing. GTD Done now completes one frozen thread plan instead of chaining best-effort label and archive calls.

## Usage

No usage change.

Closes #407.

---

### DIFF
--- a/backend/migrations/0051_folder_backfill_incomplete.sql
+++ b/backend/migrations/0051_folder_backfill_incomplete.sql
@@ -1,0 +1,26 @@
+ALTER TABLE folders
+  ADD COLUMN IF NOT EXISTS backfill_incomplete BOOLEAN NOT NULL DEFAULT false;
+
+ALTER TABLE messages
+  ADD COLUMN IF NOT EXISTS metadata_complete BOOLEAN NOT NULL DEFAULT true;
+
+-- Rows created from UID-only FETCH records before the envelope guard have this exact
+-- manufactured shape. Hide and re-fetch them; a legitimate all-NIL envelope is marked
+-- complete by the guarded upsert after verification.
+UPDATE messages
+   SET metadata_complete = false
+ WHERE message_id IS NULL
+   AND subject = '(no subject)'
+   AND COALESCE(snippet, '') = '';
+
+-- Every existing folder starts pending an exact UID verification pass. Complete rows remain in
+-- the diff set, so only legacy hollow rows (metadata_complete=false) are fetched again. Runtime
+-- backfill clears this marker without fetching for intentional provider skip views.
+UPDATE folders f
+   SET backfill_incomplete = true,
+       total_count = (SELECT COUNT(*) FROM messages m
+                       WHERE m.account_id = f.account_id AND m.folder = f.path
+                         AND m.is_deleted = false AND m.metadata_complete = true),
+       unread_count = (SELECT COUNT(*) FILTER (WHERE NOT m.is_read) FROM messages m
+                        WHERE m.account_id = f.account_id AND m.folder = f.path
+                          AND m.is_deleted = false AND m.metadata_complete = true);
--- a/backend/migrations/0052_folder_observation_generation.sql
+++ b/backend/migrations/0052_folder_observation_generation.sql
@@ -1,0 +1,68 @@
+ALTER TABLE folders
+  ADD COLUMN IF NOT EXISTS observation_generation BIGINT NOT NULL DEFAULT 0,
+  ADD COLUMN IF NOT EXISTS is_present BOOLEAN NOT NULL DEFAULT true,
+  ADD COLUMN IF NOT EXISTS topology_identity UUID NOT NULL DEFAULT gen_random_uuid();
+
+ALTER TABLE email_accounts
+  ADD COLUMN IF NOT EXISTS mailbox_topology_generation BIGINT NOT NULL DEFAULT 0;
+
+-- Bind legacy snoozes while the last known folder/message identities are still actionable.
+-- The quarantine below deliberately invalidates every historical folder observation, so this
+-- exact, unambiguous association must happen before that reset. Headerless, ambiguous, or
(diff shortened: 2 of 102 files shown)


salmonumbrella:
@maathimself this fixes a brutal bug, I hope you can prioritize it 🙏

maathimself:
Thanks for the enormous amount of work here, and for filing #407 with such a clear problem statement. I did a deep pass on this, including adversarial tracing of the mutation, send, and migration paths, and I want to be upfront that the core engineering is genuinely strong:

- The move/copy/delete fencing holds up. The three-layer identity check (DB snapshot UIDVALIDITY, the atomic id/uid/folder/uidvalidity/generation re-check under row lock, and the live IMAP SELECT-epoch check) plus the reissued topology_identity means I could not find a way to act on the wrong message, even across a delete-and-recreate that reuses UIDVALIDITY. The operation_key + travelling marker + advisory lock make retries genuinely idempotent, and reconciliation correctly refuses to guess.
- The send/forward path is safe: provider_started is durable before SMTP DATA, uncertain/started rows are never auto-retried, and the legacy pending -> uncertain reclassification is the right call. No duplicate-send or false-sent is reachable.

So this is not a quality rejection. But I can't merge it in this form, for one concrete reason and a couple of smaller ones.

The blocker is the deploy behavior. Migration 0052 nulls uid_validity on every folder and marks every message metadata_complete=false in one transaction, and since every read surface gates on metadata_complete, the entire mailbox goes dark the moment the migration runs. Recovery then deletes the quarantined rows in seedFolderUidValidity and re-FETCHes every message from the provider via backfill. On my own instance that's ~99k messages across 8 accounts (including iCloud, which throttles hard), so deploying this means hours of empty or partial folders, unread badges reading zero, and every provider getting a full-history refetch. For a change whose observable target is transient ghost rows and some GTD-Done race edges, a guaranteed multi-hour full re-download on every upgrade is too much blast radius.

Two related things I'd want addressed:
- Snoozes: due snoozes can't fire during the recovery window, and ambiguous ones end up permanently orphaned (left resolution_state='active', never fired, never surfaced for the manual reconciliation the comment describes).
- Retention: raw RFC822 and recipients are correctly redacted on successful send, but a send or forward stranded in uncertain/provider_started keeps full plaintext at rest indefinitely, with no sweeper.

What I'd ask before this can land:
1. Rework the migration so it does not blanket-quarantine and re-download. Verify existing rows in place, or quarantine lazily and only for folders whose identity is genuinely unconfirmed, so a normal upgrade doesn't dark the mailbox or refetch everything.
2. Split this into independently reviewable and revertible PRs. The send-idempotency layer, the move/delete fencing, the GTD-Done frozen-plan work, and the ghost-row visibility fix each stand on their own. A single 20k-line commit is impossible to bisect if something misbehaves in production, and much of it is mergeable well before the rest.
3. Add cleanup for stuck uncertain/provider_started operations and their retained payloads.

I'm keeping the PR open. If you're up for the migration redesign and the split, I'll review the pieces as they come, and the fencing/send work in particular I'd be glad to take early once it's on its own branch.

maathimself:
I've addressed #407 directly in bf0d47a with a small, targeted change to the move-reconciliation path (plus ghost-row suppression), verified end-to-end against the live scenario. Given that, I'm closing this PR rather than merging it. 

There are two reasons:

- The migration re-quarantines and re-downloads the entire mailbox on deploy (every folder loses UIDVALIDITY, every message is marked incomplete until a full re-fetch). On a large account that's a multi-hour dark window, too much blast radius for the problem.
- It's a single ~20k-line commit spanning the whole mutation/sync/send surface, which can't be safely reviewed or bisected as one unit.

Thank you.

Issue #402, opened 2026-08-24
Move-to-folder pickers: show ancestor paths and search the full path


Monkey7539:
**Is your feature request related to a problem? Please describe.**
In every move-to-folder picker (context menu, message pane picker, mobile move sheet, bulk-move picker), folders are listed by bare name only. With nested folders this makes same-named folders indistinguishable: an account with `INBOX/Archive` and a root `Archive`, or `Personal/Insurance` and `Business/Insurance`, shows two identical `Archive` / `Insurance` rows and the user has to guess which one they're moving mail into. The search box has the same limitation — it only matches the folder's own name, so there's no way to narrow by parent (typing `personal` finds nothing that helps you pick the right `Insurance`).

**Describe the solution you'd like**
Two complementary changes, sharing one small delimiter-aware util:

1. **Ancestor-path labels**: each picker row shows the folder's parent chain as a muted prefix before the name — e.g. `Personal / Insurance` with `Personal /` in tertiary color and `Insurance` in the normal text color. Root folders are unchanged. The parent portion truncates first when space is tight, so the folder name itself stays readable. Applies to all four pickers (including their Recent/Favorites sections) via one shared label component.
2. **Path-aware search**: the picker search matches any path segment, and accepts `parent/child` queries normalized across account delimiters (a `/` in the query matches accounts whose IMAP delimiter is `.`, e.g. Dovecot's `INBOX.Receipts.Amazon`).

As optional follow-up polish (separate PR so it can be taken or left independently): labels that still truncate can slide left marquee-style on hover to reveal the clipped remainder, snapping back on mouse leave.

**Describe alternatives you've considered**
- Showing the full path as one plain string: with end-ellipsis, deep paths truncate the folder name itself — the most important part — which is why the two-tone prefix + parent-truncates-first layout is preferable.
- A hierarchical tree inside the picker: heavier UI, conflicts with the flat Recent/Favorites sections, and makes search results awkward.
- Tooltips with the full path: invisible on touch devices and slower to scan than an always-visible prefix.

**Additional context**
I have both changes implemented and running in production on my instance (with unit tests for the path/search util, delimiter-aware for `/` and `.` accounts). Happy to submit as two PRs — labels + search first, hover-scroll as an optional second — if this direction works for you. Same contributor as #394–#398.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

maathimself:
[...]

PR #410, opened 2026-08-27
feat(move-picker): show ancestor paths and search the full path

Monkey7539:
## Summary

Implements the ancestor-path labels + path-aware search approved in #402. Folders with the same name under different parents (`INBOX/Archive` vs a root `Archive`, `Personal/Insurance` vs `Business/Insurance`) were indistinguishable in every move-to-folder picker, and the search box only matched the folder's own name.

Each picker row now shows the folder's ancestor chain as a muted prefix before the name — `Personal / Insurance` with the parent portion in tertiary color — via a shared `FolderPathLabel` component. Root folders are unchanged. When space is tight the parent truncates first so the folder name stays readable. The picker search matches any path segment and accepts `parent/child` queries normalized across account delimiters (a `/` in the query matches `.`-delimited accounts like Dovecot's `INBOX.Receipts.Amazon`).

## Changes

- New `frontend/src/utils/folderDisplay.js`: `folderDelimiter`, `folderParentPath`, `folderParentLabel`, `folderMatchesQuery`. Per your ask in #402, the delimiter/parent primitives live here as the single source of hierarchy: `sidebar.js`'s tree builder now imports `folderDelimiter`/`folderParentPath` from this module instead of keeping its own copies, so the sidebar tree and the pickers can never diverge.
- New `frontend/src/components/FolderPathLabel.jsx`: the two-tone label, used by **all** pickers — context menu (including Recent and Favorites sections), message pane picker, mobile move sheet, and the bulk-move picker. Falls back to `folder.path` when a folder has no name.
- Search filters in all four pickers switch from name-only `includes` to `folderMatchesQuery`.
- Unit tests (`folderDisplay.test.js`, 18 cases) covering `/` and `.` delimited accounts, per-folder delimiter fallback, and the cross-delimiter edge cases you asked for: a `/`-account folder named `2024.05` does **not** false-match the query `2024/05`, and a `.`-account folder literally named `a/b` stays searchable by that name with its parent chain derived from the real delimiter.

## Testing

- Running in production on my instance for a week across Gmail (`/`) accounts with heavily nested folders; duplicate `Archive`/`Drafts` entries are now unambiguous in every picker, and parent-segment searches narrow as expected.
- `npm run lint` clean; full frontend suite passes (1686/1686) including the new util tests and the sidebar suite, which now exercises the shared primitives through the tree builder.

Optional follow-up (hover-scroll for labels that still truncate) is stacked separately as discussed in #402.

---

### DIFF
--- a/frontend/src/components/ContextMenu.jsx
+++ b/frontend/src/components/ContextMenu.jsx
@@ -5,6 +5,8 @@ import { api } from '../utils/api.js';
 import { getContextMenuPolicy, resolveContextMenuMessage } from '../utils/contextMenuPolicy.js';
 import { usePluginCollected } from '../plugins/PluginSlot.jsx';
 import MessageHeaderModal from './MessageHeaderModal.jsx';
+import FolderPathLabel from './FolderPathLabel.jsx';
+import { folderMatchesQuery } from '../utils/folderDisplay.js';
 import { useUiScale, descale } from '../hooks/useUiScale.js';
 import { useMobile } from '../hooks/useMobile.js';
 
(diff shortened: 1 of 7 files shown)


maathimself:
Merged in 0f47f06 (rebased onto main, so GitHub didn't auto-close this). Both asks met: the delimiter primitives moved into folderDisplay.js with sidebar.js importing them, and the tests cover dot-delimited accounts plus the cross-delimiter false-match cases. Thanks.

PR #411, opened 2026-08-27
feat(move-picker): hover-scroll truncated folder path labels

Monkey7539:
## Summary

Optional follow-up to the ancestor-path labels from #402: a deep path like `Ice Rentals / Groups / Beacon Hill` can still truncate in a 320px picker, leaving no way to read the clipped part. On hover, a label that overflows slides left marquee-style to reveal the remainder, then snaps back on leave. Labels that fit are completely unaffected, and there is no motion on touch devices (no hover).

**Rebased onto main** now that the labels PR has landed (0f47f06), so this is a single commit that only touches `FolderPathLabel.jsx`. Still deliberately separate so it can be dropped if the motion isn't wanted.

## Changes

- `FolderPathLabel` gains hover measurement: while hovered the segments stop flex-shrinking (natural width) so the true overflow can be measured, then the content translates left by exactly the clipped amount. The reveal pace scales with distance (~80px/s, min 0.4s) and a 0.3s delay keeps quick mouse passes over the list from triggering slides.
- Pure CSS transform animation; no timers or rAF loops. State resets on mouse leave with a short ease-out.

## Testing

- Running in production on my instance: deep three-level paths reveal fully on hover in the context menu, message pane picker, and bulk-move picker; short labels show no motion; sweeping the cursor down the folder list triggers nothing.
- On the rebased branch: `npm run lint` clean, `npm run build` succeeds, full frontend suite passes (1913/1913), and `npm audit --omit=dev --audit-level=high` passes.

---

### DIFF
--- a/frontend/src/components/ContextMenu.jsx
+++ b/frontend/src/components/ContextMenu.jsx
@@ -5,6 +5,8 @@ import { api } from '../utils/api.js';
 import { getContextMenuPolicy, resolveContextMenuMessage } from '../utils/contextMenuPolicy.js';
 import { usePluginCollected } from '../plugins/PluginSlot.jsx';
 import MessageHeaderModal from './MessageHeaderModal.jsx';
+import FolderPathLabel from './FolderPathLabel.jsx';
+import { folderMatchesQuery } from '../utils/folderDisplay.js';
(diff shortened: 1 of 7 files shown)


maathimself (approved):
lgtm

PR #414, opened 2026-09-01
feat(rules): show ancestor paths in the move-destination dropdown

Monkey7539:
## Summary

Companion to #410, applying the same ancestor-path treatment to the inbox-rules form. The rule editor's move-destination dropdown listed bare folder names, so an account with `INBOX/Archive` and a root `Archive` (or `Personal/Insurance` vs `Business/Insurance`) showed identical entries and picking the right destination was guesswork. The rules list also summarized any move rule as just "Move", with no destination.

**Stacked on #410** — the first commit here is that PR's commit (this reuses its `folderParentLabel` util so the rules form, the pickers, and the sidebar tree all derive hierarchy from one source). Only the `AdminPanel.jsx` commit is new; happy to rebase once the base lands.

## Changes

- The move-destination `<select>` labels each option with its ancestor chain: `Ice Rentals / Groups / Beacon Hill` instead of `Beacon Hill`, via the shared `folderParentLabel`.
- The closed control gets a `title` tooltip with the selected folder's full path (a native `<option>` cannot host the pickers' two-tone label, but the opened dropdown already shows the full text untruncated).
- The rules list's action summary shows the move destination (`Move → Ice Rentals/Groups`) so rules are tellable apart at a glance.

## Testing

- Running in production on my instance: creating and editing move rules against a Gmail account with duplicate folder names under different parents — destinations are now unambiguous in the dropdown and visible in the rules list.
- `npm run lint` clean; full frontend suite passes (1686/1686).

---

### DIFF
--- a/frontend/src/components/AdminPanel.jsx
+++ b/frontend/src/components/AdminPanel.jsx
@@ -30,6 +30,7 @@ import DiagnosticsReportModal from './DiagnosticsReportModal.jsx';
 import { getEffectiveShortcuts, getGroupedActions, ACTION_DEFS, SPECIAL_KEY_LABELS, parseModKey, modLabel } from '../utils/defaultShortcuts.js';
 import { unifiedUnreadTotal } from '../utils/unifiedInbox.js';
 import { isValidForwardAddress } from '../utils/ruleActions.js';
+import { folderParentLabel } from '../utils/folderDisplay.js';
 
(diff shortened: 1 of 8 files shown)


maathimself:
Merged in 266c5ab (rebased, so GitHub didn't auto-close). Once #410 was on main this came down to 21 lines in AdminPanel.jsx, which is a nice result for the split. Thanks.

PR #419, opened 2026-09-06
fix(sidebar): make the subfolder-create input appear and stay usable

Monkey7539:
## Summary

Follow-up to #395, which fixed the backend half of subfolder creation. The UI half has a matching bug: right-clicking a folder and choosing "New subfolder" on a folder that has **no children yet** silently does nothing — no input, no error. The create input is rendered inside the children block, which only exists when the folder already has visible children and is expanded, so on a leaf folder the click sets state and nothing appears. Since every freshly created folder is a leaf, subfolders of new folders cannot be created from the UI at all.

## Changes

- Render the create input whenever creation targets the folder, outside the children block, so it appears on leaf folders too.
- Un-collapse a collapsed target folder when creation starts, so the input isn't hidden along with its children.
- Drop the input row's decorative folder icon: its footprint squeezed the text field to a sliver once the parent sits a few levels deep (indent grows 14px per level while the row width is fixed), and the indent alone already places the row among its future siblings.

## Testing

- Reproduced on a live self-hosted instance (source build) against Gmail: "New subfolder" on any childless folder was a silent no-op; after the fix the input appears indented under the target at any depth, with usable width, and creation works end-to-end with #395's backend fix.
- `npm run lint` clean; full frontend suite passes (1801/1801).

---

### DIFF
--- a/frontend/src/components/Sidebar.jsx
+++ b/frontend/src/components/Sidebar.jsx
@@ -700,6 +700,10 @@ export default function Sidebar() {
           setCreatingFolder({ accountId, parentPath: folderObj.path });
           setCreateName('');
           if (!expandedAccounts[accountId]) setExpandedAccounts(prev => ({ ...prev, [accountId]: true }));
+          // Un-collapse the target folder so the input isn't hidden with it.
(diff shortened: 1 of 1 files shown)


maathimself (approved):
lgtm

PR #420, opened 2026-09-06
fix(folders): create folders at the server's real path, not a guessed one

Monkey7539:
## Summary

Completes #395 on the server side. The create-folder route called raw `mailboxCreate` with a locally built path: no namespace handling, so on prefixed servers (e.g. everything under `INBOX.`) the folder was created at a different real path than the DB stored — a ghost row pointing at a path the server never had. The row was also inserted with a NULL delimiter, and the subfolder delimiter lookup was an unordered `LIMIT 1` that could land on such a NULL row and fall back to `/` on a dot-delimited server. Chained together: creating a subfolder under a MailFlow-created folder built a wrong-format path, and the next folder-structure sync pruned it — a silent failure.

## Changes

- Route creation through `ensureFolder` — the existing namespace/delimiter/already-exists handling the GTD plugin already relies on — joining parent and leaf with `/` for it to translate, and store the **returned real server path** in the DB.
- Store the account's delimiter on the inserted row (previously NULL, poisoning later subfolder/rename path building), resolved via a delimiter lookup that skips NULL/empty rows.
- Delete the now-unused raw `createFolder` method so nothing regresses to it.

## Testing

- New `mail.createFolder.test.js` (4 tests, modeled on the emptyFolder route tests): prefixed-server path translation (`INBOX.Foo/Bar` request → `INBOX.Foo.Bar` stored), root creation, NULL-delimiter fallback for accounts with no synced folders, and no DB insert when the IMAP create throws.
- Running in production on my instance; combined with the UI fix, subfolder creation works end-to-end including under folders MailFlow itself created.
- `npm run lint` clean; full backend suite passes (1251/1251).

---

### DIFF
--- a/backend/src/routes/mail.createFolder.test.js
+++ b/backend/src/routes/mail.createFolder.test.js
@@ -1,0 +1,86 @@
+import { afterAll, beforeAll, beforeEach, describe, expect, it, vi } from 'vitest';
+
+vi.mock('../services/db.js', () => ({ query: vi.fn() }));
+vi.mock('../middleware/auth.js', () => ({
+  requireAuth: (req, _res, next) => { req.session = { userId: 'user-1' }; next(); },
+}));
+vi.mock('../index.js', () => ({ imapManager: { ensureFolder: vi.fn(), broadcast: vi.fn() } }));
+
+import express from 'express';
+import mailRoutes from './mail.js';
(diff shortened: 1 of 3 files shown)


maathimself (approved):
lgtm

PR #421, opened 2026-09-06
fix(search): surface search failures instead of empty or stale results

Monkey7539:
## Summary

A failed search request is swallowed: the catch only `console.error`s, so the user sees an empty list for a query that has matches — or, worse, the **previous** query's results lingering under the new query text. The most common trigger is the search rate limiter itself: 20 requests/minute, while search-as-you-type fires a debounced request per typing pause, so a user exploring a few queries hits 429s that then read exactly like "no results". This makes search feel randomly unreliable (it cost me a while to diagnose on my own instance).

## Changes

- On a failed search: clear the stale results and show a "Search failed" empty state with the server's reason and a Retry button, instead of a misleading "No results".
- Raise the rate limit from 20 to 60/minute, sized for debounced typed queries (the limiter itself stays).
- New `messageList.searchFailed` key in all ten locales.

## Testing

- Reproduced on a live self-hosted instance: rapid exploratory searching hit the limiter and showed empty/stale results with no feedback; after the fix the failure state appears with the Retry-After message and Retry re-runs the query.
- `npm run lint` clean; full frontend suite passes (1802/1802) including the i18n key-parity tests.

---

### DIFF
--- a/backend/src/routes/search.js
+++ b/backend/src/routes/search.js
@@ -6,7 +6,9 @@ import { resolveAccountScope } from '../services/unifiedInbox.js';
 const router = Router();
 router.use(requireAuth);
 
-// Simple in-memory rate limiter: 20 searches per minute per user.
+// Simple in-memory rate limiter: 60 searches per minute per user. Typed search
+// fires a debounced request per pause, so a user exploring several queries can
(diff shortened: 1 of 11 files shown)


maathimself (approved):
lgtm

PR #425, opened 2026-09-07
fix(attachments): keep inline-disposed named text files out of the body

Monkey7539:
## Summary

An attached `.html` (or `.txt`) file that the sender marks `Content-Disposition: inline` — or leaves without a disposition — never appears in the attachment list and cannot be downloaded, while the message body itself renders fine. `walkStructure` routes every `text/html` / `text/plain` leaf into the body candidates before ever looking at the filename, so such a part is absorbed as a spare body part and dropped; only parts with an explicit `attachment` disposition escape that. Whether a given email is affected depends purely on how the sending client labels the part, which makes it look intermittent.

## Changes

- Text leaves now record their filename (from the disposition or the Content-Type `name` param). After the walk, a post-pass reclassifies **named** text parts as attachments — but only when an **unnamed** text part exists to serve as the body. If every text part is named (some clients name their body parts), nothing changes, so ordinary rendering can't regress.
- `walkStructure` becomes a thin wrapper (walk + post-pass) around the recursion, so all five consumers — sync-time body extraction, on-demand body fetch, attachment lookups — get the same classification with no call-site changes.
- `detectAttachments` (the paperclip flag) mirrors the rule so the icon agrees with the list.

## Testing

- Reproduced on a live self-hosted instance with a forwarded email carrying an inline-disposed HTML attachment: the body rendered but the attachment list was empty; after the fix the file lists and downloads. Messages with attachment-disposed files and plain multipart/alternative bodies behave exactly as before.
- New `walkStructure` cases: inline-disposed named HTML alongside a body → attachment; Content-Type-named text part → attachment (encoding preserved for download decoding); all-named text parts → body. New `messageParser.attachments.test.js` (5 cases) covers the paperclip rule.
- `npm run lint` clean; full backend suite passes (1255/1255).

---

### DIFF
--- a/backend/src/services/imapManager.js
+++ b/backend/src/services/imapManager.js
@@ -553,10 +553,33 @@ function decodeAttachmentBuffer(buf, encoding) {
 }
 
 export function walkStructure(node, results) {
+  walkNode(node, results);
+  // A text part that carries a filename is an attached file (an .html report, a
+  // .txt log) when the message also has an unnamed text part serving as its
+  // body. Senders often mark such files inline or omit the disposition, so the
+  // disposition check alone absorbed them into the body candidates and they
+  // vanished from the attachment list. When every text part is named, leave
+  // them as body — some clients name their body parts.
+  const named = results.textParts.filter(p => p.filename);
+  if (!named.length || named.length === results.textParts.length) return;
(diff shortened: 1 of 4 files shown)


maathimself (approved):
lgtm

PR #426, opened 2026-09-07
fix(rules): run the inbox rules sweep in the background

Monkey7539:
## Summary

"Run rules on inbox" swept every INBOX message synchronously and only answered when finished. On a large mailbox that outlives the proxy timeout (60s in the bundled nginx), so the browser receives a 504 and the UI reports failure — while the sweep keeps running server-side, applying moves the user was just told didn't happen. Retrying starts a second concurrent sweep over the same messages.

## Changes

- `POST /rules/run` now responds **202 immediately** and runs the sweep in the background, mirroring the existing `/folders/empty` route. One run per user at a time — a concurrent request gets `409 Rules are already running` instead of a duplicate sweep.
- Completion is delivered over a new `rules_run_complete` WebSocket event carrying `{ ok, processed, matched }` (or `ok: false` if the sweep itself threw). The sweep body is extracted into `runRulesSweep()` unchanged; per-account failures are still logged and skipped.
- Client: `useWebSocket` toasts the outcome, re-dispatches the existing `mailflow:rules-ran` refresh (rules move messages between folders), refreshes unread counts, and forwards the result to the settings panel via a window event. The Rules tab shows a "Running in the background — you'll be notified when it finishes" note until the result arrives, and surfaces the 409 message if a run is already active.
- New `admin.rules.runStarted` key in all nine locales.

## Testing

- Reproduced on a live self-hosted instance (~15k INBOX messages): the button previously flipped to an error after 60s while server logs showed the sweep continuing for minutes. After the fix the click returns instantly, the running note shows, and the completion toast arrives with the matched/processed totals; a second click during the run is rejected cleanly.
- New `rules.run.test.js` (3 cases, modeled on the emptyFolder route tests): immediate 202 plus a completion broadcast with correct totals scoped to the requesting user; 409 while in flight then accepting again once done; zero-rules completion with no rule evaluation.
- `npm run lint` clean on both sides; backend 1250/1250, frontend 1802/1802 (including locale-parity tests).

---


maathimself (approved):
lgtm

PR #430, opened 2026-09-09
fix(body): stop rendering attachment-only messages as decoded binary

Monkey7539:
## Summary

A message whose only MIME part is an attachment renders as decoded binary in the message pane. The everyday case is a DMARC aggregate report from `noreply-dmarc-support@google.com`: the entire message is one `application/zip` part with `Content-Disposition: attachment` and no text body. The attachment card is correct, but the body area shows `PK` followed by a wall of garbage — the zip's bytes.

Cause: both body extractors (`extractBodyFromMsg` at sync time and `fetchMessageBody` on open) have a fallback for the no-text-part case that promotes the root part to `text/plain` *regardless of its type*, so the zip's base64 is decoded and served as the message. The same happens for a `multipart/mixed` that holds only a single file — the fallback grabs part `1`, which is that file.

## Changes

- Add `bodyFallbackApplies(results)`: the fallback now runs only when the walk filed **no** attachments. If it found attachments and no text, the message simply has no body and `html`/`text` stay `null` — the pane shows the attachment card over an empty body, which is the honest rendering. A bare unrecognized single part with no attachment classification still falls back exactly as before.
- Applied in both extractors; `extractBodyFromMsg` is now exported for testing.

## Testing

- Reproduced on a live self-hosted instance with Google's daily DMARC reports to a `postmaster@` address: body showed decoded zip bytes; after the fix the message shows the attachment card and an empty body, and the zip downloads normally. (Rows cached before the fix need their `body_text`/`snippet` cleared once to re-render.)
- Three new tests: single-part zip root yields null body with the attachment listed; a `multipart/mixed` holding only a file does not fall back; a bare `text/plain` root still falls back to text.
- Rebased onto current main: `npm run lint` and `lint:plugins` clean, full backend suite passes (1364/1364), and `npm audit --omit=dev --audit-level=high` passes. The earlier red run failed only at the audit step on the @tiptap/core advisories, which main has since patched.

---


maathimself:
Merged in 01486cf (rebased onto main, so GitHub didn't auto-close this).

The CI failure was the stale base, not the change — it's green on current main, and #425 landing cleared the conflict. Nice catch on the fallback only being meant for a bare unrecognized root; gating it on the walk having found no attachments is the right fix. Thanks.

PR #444, opened 2026-09-14
fix(gtd): keep waiting labels until explicit clear

### LINKED ISSUE #443: Keep Watch and Delegated labels until explicitly cleared
## The situation

Watch and Delegated are waiting states. They should stay on a thread until the user explicitly clears them or marks the thread done.

The shared reply transition currently treats both labels like reply-cleared states. An incoming reply removes Watch, and an outgoing reply removes Delegated. That makes waiting threads disappear from their queues even though the user never cleared them.

## Ask

Keep Watch and Delegated across incoming mail, sent replies, drafts, and reconciliation. Preserve the existing reply behavior for Todo and Someday, keep Reference unchanged, and leave manual label removal and explicit done untouched.

salmonumbrella:
## What changed

- Watch and Delegated now persist across incoming mail, owner replies, drafts, and reconciliation until the user removes the label or marks the thread done.
- Todo and Someday still clear only when the latest reply is a non-draft message sent by the owner. Incoming replies do not clear them.
- Reference stays unchanged, and the transition remains shared across providers.
- Regression coverage now locks down incoming, outgoing, draft, and repeated reconciliation behavior.

## Why

Watch and Delegated are waiting states, but replies were treating them like completed work. Threads could disappear from their waiting queues without the user explicitly clearing them.

## Usage

No usage change.

Closes #443

---


maathimself (approved):
lgtm

PR #446, opened 2026-09-15
fix(sidebar): prevent conjoined highlight and add folder spacing

facundopignanelli:
## Summary

Fixes an issue where the account header row and the "INBOX" folder row visually touch and appear conjoined with a double-highlight when an account is expanded.

## Changes

- **Sidebar active state**: In `frontend/src/components/Sidebar.jsx`, only set `isAccountActive` when the account folder list is collapsed (`sidebarCollapsed || !expanded`). When expanded, the account row acts as a header and expander, while the dedicated "INBOX" folder row below it receives the active highlight.
- **Visual spacing**:
  - Added `margin: '1px 0'` to the account row header.
  - Added `style={{ marginTop: 2 }}` to the folder tree container below the account header.
  - Added `margin: '1px 0'` to folder rows (`renderNode`), the create folder input row, the "+ New folder" button, and the hidden folder toggle button to prevent edge-to-edge touching on hover or selection.

## Testing

- Verified all 1,942 frontend tests pass (`npm test`).
- Verified linter passes with 0 warnings (`npm run lint`).

---


maathimself:
One thing I'd like before merging.

The change assumes an INBOX row always exists to take the highlight when the account is
expanded. There are two cases where it doesn't, and in both the highlight disappears
entirely rather than moving down:

1. INBOX can be hidden. The hide menu item has no `isProtected` guard, unlike rename and
   delete which both carry `disabled: isProtected`, and `renderNode` returns null for a
   hidden folder. So INBOX can be absent from the tree while still being the selected
   folder.

2. The first expansion of an account fetches its folders asynchronously (`toggleAccount`
   flips `expanded` immediately and the fetch lands later), so there's a window where the
   tree is empty and no INBOX row exists yet. Brief, and only on first expand, but visible.

Both have the same fix: fall back to highlighting the account row when no INBOX row is
actually rendered, rather than keying only on `expanded`. Heads up that the values you'd
need for that (`accountFolders`, `accountHiddenPaths`, `showingHidden`) are currently
declared below where `isAccountActive` sits, so it likely means hoisting them.

facundopignanelli:
Thanks for the catch and detailed explanation!

I've updated the logic to fall back to highlighting the account row whenever `selectedFolder === 'INBOX'` and no INBOX row is actually rendered in the tree.

- Hoisted `accountFolders`, `accountHiddenPaths`, and `showingHidden` to the account mapping level in `Sidebar.jsx`.
- Evaluated whether an INBOX row is actually rendered via `hasRenderedInbox(accountFolders, { expanded, sidebarCollapsed, hiddenPaths, showingHidden })`, covering both hidden INBOX folders and the initial asynchronous folder fetch window.
- Cleaned up the duplicate inner declarations of `accountHiddenPaths` and `showingHidden`.
- Added unit tests in `sidebar.test.js` covering all cases (rail collapsed, tree folded, visible INBOX, async/unloaded folders, hidden INBOX when hidden folders are omitted, and hidden INBOX when toggled visible). All 1,948 frontend tests pass and linter is clean.

Monkey7539 (commented):
Thanks for turning the edge cases around so quickly. The `hasRenderedInbox` fallback reads well, and the tests cover every case raised above.

One suggestion on the spacing. With `margin: '1px 0'` on folder rows, adjacent row boxes sit 1px apart, and that band belongs to no element. Folder rows are drop targets, both for dragged messages and for folder reordering, so a drag down the list crosses a pixel with no `dragover` handler at every boundary: the cursor flips to "cannot drop", and a release that lands exactly there does nothing.

A transparent bottom border gives the same visible gap while keeping that band inside the row. In the folder row's style, in place of the margin:

```jsx
borderBottom: '1px solid transparent',
cursor: isRenaming ? 'default' : 'pointer',
backgroundColor: (msgDragTarget === `${account.id}:${folder.path}`) ? 'var(--accent-dim)' : isFolderSelected ? 'var(--bg-hover)' : 'transparent',
backgroundClip: 'padding-box',
transition: 'background-color 0.1s',
```

and in the same row's hover handlers:

```jsx
onMouseEnter={e => { if (!isFolderSelected && !isRenaming) e.currentTarget.style.backgroundColor = 'var(--bg-tertiary)'; }}
onMouseLeave={e => { if (!isFolderSelected) e.currentTarget.style.backgroundColor = 'transparent'; }}
```

The switch to `backgroundColor` is the part that is easy to miss. The `background` shorthand resets `background-clip`, so if it stays, in the style object or in the handlers, the highlight paints straight over the transparent border and the gap disappears again.

The other places you added spacing (the account header, the tree container, the create-folder input and the two buttons) are not drop targets, so their margins are fine as they are.

Separately, while testing this I found a pre-existing seam at fractional display scaling, where the wrapper div around each folder row wins the hit test during a drag. It is unrelated to your change, so I have sent it on its own as #456 rather than asking you to take it on.

maathimself (approved):
lgtm

PR #451, opened 2026-09-16
feat(attachments): warn before downloading a risky file type

### LINKED ISSUE #450: Attachment cards give no warning before downloading a dangerous file type

An attachment card shows a filename, a size and a paperclip, and nothing else. `invoice.pdf.exe` looks exactly like `invoice.pdf` until it is on disk, and a macro-enabled `.docm` looks exactly like a `.docx`. The declared MIME type is no help — it is chosen by the sender, and the usual carriers arrive as `application/octet-stream` regardless of what they really are.

This is the one place in the client where a single click hands an attacker's file to the OS, and it is also the place where the UI says the least.

**Describe the solution you'd like**

Classify each attachment by its real extension and say so on the card, with the download itself gated for the risky ones:

1. **Dangerous** (red border, bold): executables, scripts, installers, disk images and shortcuts — `.exe .scr .com .pif .bat .cmd .ps1 .vbs .js .jar .msi .dmg .lnk` and similar. "Dangerous file type (.exe) — can run code on your computer".
2. **Risky** (amber): macro-enabled Office files and standalone `.html`/`.svg` pages, which carry macros or a fake login form. "Risky file type (.docm) — may contain macros or a fake login page".
3. **Archive** (muted): "Archive — contents cannot be inspected", because a `.zip` is neither safe nor unsafe on its own and the client cannot see inside it.
4. **Double extension** called out by name: `invoice.pdf.exe` reads "Hidden extension: .pdf.exe — the real type is the last one", since that disguise is the whole trick.

Downloading a red or amber file takes a second click. The first click arms the button and shows the reason, so the warning is read rather than dismissed. Anything unclassified behaves exactly as it does today, with no extra click and no badge.

The classifier is a pure function (`frontend/src/utils/attachmentRisk.js`) with its own unit tests. The declared MIME type is consulted only as a fallback for a file with no extension at all. Nothing is blocked outright and nothing is sent anywhere — the client already has the filename, so this is display plus one confirmation.

**Describe alternatives you've considered**

- **Scanning attachment content** (ClamAV, a hash lookup against a reputation service): far stronger, but it means a new dependency or an outbound call per attachment, a daemon to run, and a privacy decision that belongs to the operator rather than to me. Extension policy is what a client can do honestly on its own.
- **Blocking dangerous types outright**: wrong for a self-hosted client. People do legitimately mail each other installers, and a client that refuses is a client people work around.
- **Leaving it to the mail server**: gateways do strip many of these, but plenty of self-hosted setups have no such filter, and the classifier in #388 scores the message rather than the file.
- **Warning only on the double extension**: catches the obvious disguise and misses the plain `.exe` that people still open.

**Additional context**

Running on my own instance. This does not overlap #388: that scores whether a message is spam from its content, this says what a particular file is, and the two disagree usefully — a targeted attachment arrives in a message that looks entirely ordinary.

Happy to open the PR if the direction works for you: 177 lines, frontend only, no new dependency, nine locales, full suite green. Same contributor as #394-398 / #410 / #411 / #414 / #419-422 / #425 / #426 / #430 / #442.

Monkey7539:
## Summary

Implements the attachment risk badges discussed in #450. An attachment card shows a filename, a size and a paperclip, so `invoice.pdf.exe` looks like every other attachment until it is on disk, and a macro-enabled `.docm` looks like a `.docx`. The declared MIME type cannot settle it — the sender chooses it, and the usual carriers arrive as `application/octet-stream` either way.

Each attachment is now classified by its real extension, the card says what the file is, and a risky one takes a second click to download. Nothing is blocked, and anything unclassified looks and behaves exactly as it does today.

## Changes

- New `frontend/src/utils/attachmentRisk.js`: `classifyAttachmentRisk(filename, mimeType)` returns `{ level, ext, doubleExt }` where level is `block`, `warn`, `notice` or `ok`.
  - `block` — executables, scripts, installers, disk images and shortcuts.
  - `warn` — macro-enabled Office files and standalone `.html`/`.svg` pages, which carry macros or a fake login form.
  - `notice` — archives, whose contents the client cannot inspect.
  - A disguising double extension (`invoice.pdf.exe`) is reported separately, because that is the part worth naming in the UI. It only counts when the fake half is a document or media type, the same list the backend `ATTACHMENT_DOUBLE_EXT` rule uses, so dates and version numbers in a filename are not flagged.
  - The MIME type is consulted only as a fallback for a file with no extension at all.
- `MessagePane` colours the card border and prints the reason under the filename. A `block` or `warn` file needs a second click: the first arms that one card and shows why, and moving to another message clears it.
- Nine locales get `message.attachmentRisk.*`, filed beside the other attachment strings.

## Testing

- `attachmentRisk.test.js`, 8 cases: each level from the extension, the MIME fallback, a double extension, dotted dates and version numbers (`Statement 09.15.2026.html`, `P&L v2.1.xlsm`) not being reported as hidden extensions, an uppercase extension, and an ordinary document staying `ok`.
- Running on my own instance: ordinary attachments (images, PDFs) carry no badge and still download on the first click, so the quiet path is confirmed against real mail. The `block` and `warn` paths are covered by the unit tests above rather than by a live sample.
- `npm run lint` clean, `npm run build` succeeds, full frontend suite passes (2026/2026), and `npm audit --omit=dev --audit-level=high` passes.

Closes #450

---


rafpigna:
Hello @Monkey7539  this is real and something that I never tought to cover in my antispam classifier #388   because, as you already noticed,  it is focused on the message content, never reaching attachments. 

I just want to say that I'm available and really happy to help in any way if any of you think that this should be managed or used in some way by the antispam engine. 

I mean: your #451 adds the warning, the antispam engine could add an auto-move to spam folder and training based on user feedback considering also the attachments, since now is totally ignored by the training and by the auto classification/move. 

Anyway, I'm unsure that the spam classifier/training/automove should involve the attachments in some way, so about this I'm totally open on what @maathimself wants and to help in any way.

Monkey7539:
Thanks @rafpigna, and thanks for offering to help.

I think #451 should stay display-only. It's a safety check at the moment you open or download a file, a warning plus a second click, rather than a spam signal. "Dangerous attachment" and "unwanted mail" only partly overlap, and the amber and archive tiers deliberately over-warn on everyday mail like HTML invoices and DMARC report zips. That's fine for a warning you can click past, but it isn't something I'd want feeding a spam score or an auto-move.

While looking into it I read through the antispam code on main, and the engine turns out to already have attachment signals (the `ATTACHMENT_EXECUTABLE` / `ATTACHMENT_DOUBLE_EXT` rules and the attachment flag tokens). From the code, though, they don't seem to reach the verdict on incoming mail: the attachment list is still empty when a new message is classified, and the flag tokens are trained but not used when scoring. I've written that up with line references in #457 so this PR stays focused. It comes from reading the code rather than a live database, so please correct me if I've missed something, and whether to wire those signals in or drop them is your and @maathimself's call.

Looking into this also turned up a bug on my side: #451's hidden-extension check was flagging dates and version numbers, so "Statement 09.15.2026.html" was reported as hiding ".2026.html". That's fixed in 34605ba, which now only counts a document or media type as the fake half, the same list the backend `ATTACHMENT_DOUBLE_EXT` rule uses.

maathimself (approved):
lgtm

PR #456, opened 2026-09-16
fix(sidebar): stop the folder row wrapper winning drag hit-tests

Monkey7539:
## Summary

Dragging a message onto the folder list, or dragging a folder to reorder it, flashes the "cannot drop" cursor for an instant at the boundary between two folder rows, even though the rows touch. It shows plainly on a 175% Windows desktop.

Under fractional display scaling a row's box ends on a fractional pixel, and at that seam the element winning the hit test is the wrapper `div` that `renderNode` puts around each row and its children, not either row. The wrapper has no drag handling, so the `dragover` for that sliver goes unanswered and the browser shows the refusal cursor. The wrapper has always been there, so this is independent of #446.

## Changes

- The per-folder wrapper in `renderNode` gets `pointerEvents: 'none'`, so it can no longer be the hit-test target.
- Its interactive descendants opt back in with `pointerEvents: 'auto'`: the folder row itself, and the subfolder-create input row (`createFolderInput`), which renders inside the wrapper. Nested child rows live in their own wrappers and follow the same pattern, so every row keeps its click, context menu, rename, reorder grip and drop handling.

## Testing

- Measured on a live instance at 175% display scaling by sweeping synthetic `dragover` events down the folder list in 0.2px steps and checking whether each was accepted: four sub-pixel points refused the drop. With the same `pointer-events` change applied to the running page, none did. A recording of a real folder-reorder drag showed the refusals landing on the wrapper at exactly those seams.
- `npm run lint` clean, `npm run build` succeeds, full frontend suite passes (2013/2013), and `npm audit --omit=dev --audit-level=high` passes.

---


maathimself (approved):
lgtm

PR #458, opened 2026-09-16
fix(drafts): stop draft save and discard expunging other mail

Monkey7539:
## Summary

Saving, sending or discarding a draft can permanently delete a message that isn't a draft.

Both draft routes expunge using a folder the client supplies, and check only that the account belongs to the user:

- `POST /api/mail/draft` deletes `existingUid` from `existingFolder` after saving the new copy.
- `DELETE /api/mail/draft/:uid` deletes from `?folder=`.

Neither checks that the folder is a Drafts folder, and the UI reaches that today. In a Drafts view, `handleSelect` opens every clicked row in compose as a draft carrying its own `folder`, including search results from other folders (an `in:` search, or Search all folders). Saving, sending or discarding one of those expunges the original message from wherever it lives.

On the same call, `POST` passes `existingUid` straight to `messageDelete(String(uid), { uid: true })`, which reads it as a UID set, so `"1:*"` would expunge every message in the folder. The UI never sends that, so this part is defence in depth.

## Changes

- New `isDraftsPath(account, folder, draftsFolder)` in `draft.js`. A folder counts only if it is a non-empty string and one of:
  - in the canonical Drafts set from `resolveAllDraftsPaths`;
  - the folder this file appends drafts to, since `resolveDraftsFolder` uses the raw mapping and a save must always be able to replace its own previous copy;
  - the server's own `\Drafts` folder, which `MessageList`'s `isDraftsFolder` still opens as Drafts when the mapping points somewhere else. Without this, drafts opened there would silently stop being replaced or discarded.
- `POST /draft` deletes the old copy only when `existingUid` is a single positive integer (a number, or the digit string a `BIGINT` uid comes back as) and `existingFolder` passes the check. The save itself is never refused: the new draft is still appended first, and an old copy that fails the check is left in place and logged.
- `DELETE /draft/:uid` returns `400 { error: 'Folder is not a Drafts folder' }` before touching IMAP, following the `Message is not in the spam folder` precedent in `mail.js`. All three frontend callers already `.catch(() => {})`, so a refused discard just leaves the message where it is.

Separate follow-ups, not in this PR:

- `MessageList` still opens non-draft rows in compose while a Drafts folder is selected. The server now makes that harmless, but those rows shouldn't open as drafts at all.
- Changing the From account on a reopened draft sends the new `accountId` with the old draft's uid and folder (`ComposeModal` only updates `draftAccountId` after the save succeeds), so a same-named Drafts path in the other account still passes the check.

## Testing

- 16 new cases in `draft.test.js`, using the existing harness with `mailUtils` running for real against the mocked `query`:
  - `POST`: replacing the previous copy in the Drafts folder it just saved to; a reopened draft with a string uid in a second Drafts path; a drafts mapping that isn't a synced folder, where a save must still replace its own copy; a non-Drafts `existingFolder`, where the save succeeds and nothing is expunged; and six uid values that must never expunge (`"1:*"`, `"1,2,3"`, `[1,2,3]`, `"4abc"`, `-1`, `1.5`).
  - `DELETE`: the Drafts folder; a second canonical Drafts path; a non-Drafts folder; a repeated `folder` parameter; and, with the mapping pointing elsewhere, the server's own `\Drafts` folder accepted and an unrelated folder refused.
- Against the current `draft.js`, 11 of them fail: 10 because the route expunges where it shouldn't, and 1 only because the old code passes the uid through as a string. The other 5 guard that legitimate replace and delete keep working, and pass on both.
- `npm run lint` and `npm run lint:plugins` clean, full backend suite passes (1602/1602), and `npm audit --omit=dev --audit-level=high` passes.

---


maathimself (approved):
Verified against the current code: the vulnerability reproduces, the fix closes it, and your new tests fail 11 of 21 on the vulnerable route exactly as described. The three accept cases in isDraftsPath are all load-bearing — dropping the \Drafts one would have broken legitimate draft replacement. Skipping only the expunge rather than refusing the save is the right call. Thank you for catching this.

Issue #457, opened 2026-09-16
antispam: attachment signals never reach the verdict on incoming mail


Monkey7539:
**Describe the bug**

The antispam engine has attachment signals — the `ATTACHMENT_EXECUTABLE` (0.6) and `ATTACHMENT_DOUBLE_EXT` (0.3) rules and the `__has_attachment__` / `__attachment_is_executable__` flag tokens — but as far as I can tell from reading the code, none of them can influence the verdict on incoming mail. Two separate gaps:

1. **The attachment list is empty when a new message is classified.** Every provider profile sets `fetchBody: false`, so the sync insert leaves `atts = []` and writes `attachments = '[]'`, and the classifier runs on that fresh row straight away:
   - [`imapManager.js:3148`](https://github.com/maathimself/mailflow/blob/0ab6d50e3d90f60ccff23e2c021f74d62ffab683/backend/src/services/imapManager.js#L3148) `let safeHtml = null, text = null, atts = [];` filled only under `prefetchBody && provider.fetchBody`
   - [`imapManager.js:3270`](https://github.com/maathimself/mailflow/blob/0ab6d50e3d90f60ccff23e2c021f74d62ffab683/backend/src/services/imapManager.js#L3270) `maybeClassifyNewMessage` on that row
   - [`spamPipeline.js:83`](https://github.com/maathimself/mailflow/blob/0ab6d50e3d90f60ccff23e2c021f74d62ffab683/backend/src/services/spamPipeline.js#L83) `attachments` read from `row.attachments`, so `extractFlagFeatures` ([`spamTokenizer.js:231`](https://github.com/maathimself/mailflow/blob/0ab6d50e3d90f60ccff23e2c021f74d62ffab683/backend/src/services/spamTokenizer.js#L231)) returns 0 for both attachment flags and the two attachment rules have nothing to test

   The body text is `NULL` on that row for the same reason, so automatic verdicts rest on the subject and headers. Sync already fetches `bodyStructure` (it is what `has_attachments` is derived from, [`messageParser.js:642`](https://github.com/maathimself/mailflow/blob/0ab6d50e3d90f60ccff23e2c021f74d62ffab683/backend/src/services/messageParser.js#L642)), so the attachment list could be built from it at insert without fetching bodies.

[...]

PR #459, opened 2026-09-16
fix(attachments): ask before Download all fetches a risky file

Monkey7539:
## Summary

A risky attachment needs a second click to download on its own (#451), but **Download all** above the list zips every attachment, risky ones included, on the first click with no warning ([`MessagePane.jsx#L2643`](https://github.com/maathimself/mailflow/blob/e4d8cd4cccb945921a9fbc3739881aebea681f49/frontend/src/components/MessagePane.jsx#L2643)). Raised in #457.

## Changes

When any attachment would warn or block on its own, the link asks first:

- The first click, or Enter, arms it. The link turns red and shows the existing "click again to download anyway" note. The second click starts the download.
- While it asks, the link has no `href` (it is `role="button"`), so a right-click "Save link as", a middle click, a long press or a drag has nothing to fetch around the confirm. The confirming click downloads the zip itself through a temporary link, the same way `handleDownload` saves single attachments.
- `riskArmed` holds a `DOWNLOAD_ALL` sentinel for the link. Arming a single file disarms the link and vice versa, and both reset on message change as before.

Messages with only safe attachments keep a plain `<a href download>` link, as today. No new strings.

## Testing

Four render tests in `MessagePane.render.test.js` load a message body with attachments and drive the real link. The component's own downloads are recorded by stubbing `HTMLAnchorElement.prototype.click`.

- **A blocked file (`invoice.pdf.exe`):** the link has no `href` before or after arming. The first click downloads nothing and shows the note, and the second downloads the zip once and clears the note.
- **A warn-only file (`account-login.html`):** the link asks too, and Enter arms it.
- **Switching to another message and back:** a half-confirmed link is dropped.
- **Only safe attachments:** the link has `href` and `download`, and the first click is not cancelled.

Three of the four fail against the current `MessagePane.jsx`. The warn-only test also fails if the check is narrowed to `block`.

`npm run lint` clean, `npm test` passes (2036/2036), `npm run build` succeeds.

---


maathimself (approved):
Verified. This closes the bypass properly, and dropping the `href` while it asks is the detail that makes it real: right-click "Save link as", middle click and long press all have nothing to fetch, which a plain `onClick` guard would have left open.

Confirmed the link keeps `href` and `download` when every attachment is safe, so the common case is unchanged.

Two non-blocking notes, both fine to leave for a follow-up:

1. The separator is outside the translated string: `` ` - ${t('message.attachmentRisk.confirm')}` ``. No locale can change that punctuation, and some would want different spacing or a colon. Folding it into the i18n string would be cleaner.
2. `DOWNLOAD_ALL = 'all'` relies on an attachment part never being that literal. That holds for IMAP part numbers, and your comment says so, but a `Symbol` would make it structurally impossible rather than true by convention.

Neither changes behavior, so I am merging as is. Thanks for splitting this out of #457 rather than piling it into the original PR.

PR #460, opened 2026-09-16
fix(attachments): classify names ending in a dot by their extension

Monkey7539:
## Summary

`classifyAttachmentRisk` takes the extension after the last dot, so a name ending in one has an empty extension. `invoice.exe.` is classified `ok`: no badge and no second click.

Trailing dots and spaces get dropped when the file is saved, so it can land on disk as `invoice.exe`:
- Windows drops them.
- Firefox's download sanitizer drops them too ([`test_filename_sanitize.js`](https://searchfox.org/mozilla-central/source/uriloader/exthandler/tests/unit/test_filename_sanitize.js) turns `" .whitespaceanddots.png..."` into `whitespaceanddots.png`).

Raised in #457.

## Changes

- Trailing dots, whitespace and U+180E are dropped before the extension is read. `\s` no longer matches U+180E, but Firefox's sanitizer still strips it.
- The name is classified by the extension it will be saved with, including a disguise such as `invoice.pdf.exe.`. A name that is only dots still falls back to its declared MIME type.
- The characters are stripped with a loop from the end rather than `/[.\s]+$/`. That regex backtracks quadratically on a long run of dots or spaces in the middle of a name: `'setup' + '.'.repeat(100_000) + 'exe'` took about 16 s. `classifyAttachmentRisk` runs on every render of the attachment list, so that would freeze the pane.

## Testing

- `attachmentRisk.test.js`, two new cases:
  - **Trailing characters:** `invoice.exe.`, `invoice.exe . .`, `report.docm...`, `invoice.pdf.exe.` (with its `doubleExt`), `invoice.exe` + U+180E, and `.` with `application/x-msdownload`. Fails against the current file.
  - **Speed:** names with 100,000-character runs of spaces and dots classify in under a second. They take about 2 ms here; the regex version took 16 s and fails this case.
- `npm run lint` clean, `npm test` passes (2034/2034), `npm run build` succeeds.

---


maathimself (approved):
Verified, and the backtracking note is not theoretical.

I ran `/[.\s]+$/` against `'setup' + '.'.repeat(100000) + 'exe'` here: 4.3s, against 1ms for the loop you shipped. Different machine from your 16s, same conclusion. Since `classifyAttachmentRisk` runs on every render of the attachment list, the regex version would have frozen the pane on a hostile name, so the loop is the right call and worth the comment you left explaining why.

Also checked that the strip does not cost us the disguise label: `invoice.pdf.exe.` still reports `pdf.exe`.

Thanks.

PR #461, opened 2026-09-16
fix(attachments): stop calling a converted "Letter.doc.rtf" disguised

Monkey7539:
## Summary

A double extension counts as a disguise when a document or media type sits right before a risky one (`invoice.pdf.exe`). `rtf` is in both `WARN` and `DECOY`, so a word-processing document sent as RTF under its old name, such as `Letter.doc.rtf`, shows "Hidden extension: .doc.rtf". The hidden `.rtf` opens in the same kind of program the visible `.doc` promises. Raised in #457.

## Changes

- New `RTF_SOURCES` (`doc`, `docx`, `odt`): one of these in front of `.rtf` is no longer reported as a disguise. The file keeps its normal `warn` badge.
- The exemption is deliberately narrow. A PDF, picture, media or plain-text name in front of `.rtf` (`invoice.pdf.rtf`, `holiday.jpg.rtf`, `readme.txt.rtf`) is still reported, since RTF is a common exploit carrier and those names are lures. Every other risky extension is unaffected.

## Testing

- New case in `attachmentRisk.test.js`:
  - `Letter.doc.rtf`, `Minutes.docx.rtf` and `Letter.odt.rtf` have no `doubleExt`.
  - `invoice.pdf.rtf`, `holiday.jpg.rtf`, `readme.txt.rtf` and `statement.pdf.html` keep theirs.
  
  It fails against the current file, and also against a broader rule that exempts every document type ending in `.rtf`.
- `npm run lint` clean, `npm test` passes (2033/2033), `npm run build` succeeds.

---


maathimself (approved):
Verified. The exemption is correctly narrow.

Confirmed `Letter.doc.rtf`, `Minutes.docx.rtf` and `Letter.odt.rtf` lose the disguise label while keeping the `warn` badge, so a converted document still takes a second click. `invoice.pdf.rtf`, `holiday.jpg.rtf` and `readme.txt.rtf` keep their label.

Keeping `warn` rather than dropping to `ok` is the part that makes this safe, since RTF is still a live exploit carrier. Thanks.

PR #462, opened 2026-09-16
fix(attachments): flag more types that run code when opened

Monkey7539:
## Summary

`attachmentRisk.js` classifies many well-known dangerous attachment types as `ok`, so they download on one click with no badge. Raised in #457.

## Changes

**Now `block`.** These come from [Outlook's blocked attachments list](https://support.microsoft.com/en-us/office/blocked-attachments-in-outlook-434752e1-02d3-4e90-9124-8b81e49a8519), picked because they run code, or open a program or remote location, when opened.

| Group | Types | Notes |
|---|---|---|
| Scripts | `py` `pyw` `pyz` `pyzw` `pyc` `pyo` `pl` `ksh` `csh` | `sh` is already blocked. The backend's `EXECUTABLE_EXTENSIONS` already has `pyc` `pyo` `pl` `ksh` `csh`. |
| Programs and installers | `msc` `xll` `jnlp` `app` `appref-ms` `msu` `diagcab` | Excel [blocks untrusted XLL add-ins by default](https://support.microsoft.com/en-us/topic/excel-is-blocking-untrusted-xll-add-ins-by-default-1e3752e2-1177-4444-a807-7b700266a6fb). |
| Script components | `sct` `wsc` | |
| Shortcuts | `settingcontent-ms` `search-ms` `library-ms` `website` | Like `url` and `scf`, already blocked. |

**Also `block`:** `rdp`. Remote Desktop connection files were mailed in Midnight Blizzard's October 2024 spear-phishing campaign (Microsoft Threat Intelligence).

**Now `warn`:** `xlsb`, a binary workbook that can carry macros like `xlsm`. Today only its declared `macroEnabled` MIME type catches it, so one sent as `application/octet-stream` shows nothing.

**Left out on purpose.** The rest of Outlook's list either doesn't run code or would mostly flag ordinary files:
- certificates (`cer`, `crt`, `der`);
- Access databases and shortcuts (`mdb`, `mad`…`maw`);
- legacy help and shell scrap files;
- PowerShell data and format files (`psd1`, `ps1xml`…);
- temp and profile files.

This is the frontend list only and changes no antispam scoring. Sharing one list with the backend is discussed in #457.

## Testing

- `attachmentRisk.test.js`:
  - The block case now lists all the new types and asserts the full list of misses at once, so a regression names every type it drops.
  - The warn case adds `xlsb` with a generic MIME type.
  - Against the current file, both cases fail, and the block case lists exactly the 23 new types.
- `npm run lint` clean, `npm test` passes (2032/2032), `npm run build` succeeds.

---


maathimself (approved):
Verified. Good sourcing on the list, and the exclusions are as interesting as the inclusions: leaving out certificates and Access databases avoids flagging ordinary mail.

Confirmed the hyphenated extensions classify correctly (`x.settingcontent-ms`, `x.search-ms`, `x.library-ms`), that `xlsb` warns when the declared type is generic `application/octet-stream`, and that the new types still compose with the double-extension rule (`report.doc.py` reports `doc.py`).

One judgment call rather than a defect: `.py` will be the noisiest addition here, since developers mail each other scripts and this project's own users are largely technical. A red badge and a second click is the right amount of friction for something that runs code, so I am keeping it, just noting it is the one most likely to draw a complaint.

PR #463, opened 2026-09-16
fix: message pane mark unread icon

dcoffin88:
## Summary

Update the message pane header mark-unread icon so it matches the icon used elsewhere in the app.

## Changes

- Replaced the message pane header mark-unread SVG with the same dotted-envelope icon used elsewhere.

## Testing

Tested on my dev environment.

---


maathimself (approved):
Verified, and this is worth more than the "matches the icon used elsewhere" framing.

The two glyphs carry distinct meanings here: the plain open envelope is mark read, the dotted envelope is mark unread. The ternary in `ContextMenu.jsx` picks between them on `hasUnread`. Both of the controls you changed are wired to `handleMarkUnread` and labeled `contextMenu.markUnread`, so the message pane was drawing the mark-read glyph on a mark-unread button. That is a wrong icon, not just an inconsistent one.

Checked as well:
- Those were the only two stale instances in `MessagePane.jsx`; none are left after this.
- The plain envelopes remaining in `RowHoverActions.jsx`, `ContextMenu.jsx`, `MessageList.jsx` and `AdminPanel.jsx` are all mark-read controls, so they are correctly left alone.
- Merges cleanly against current main despite #459 having just landed in this file, and the frontend suite passes.

Matching the existing `strokeMiterlimit` and inline-style pattern was the right call even though the miter limit does nothing on a circle, since all three canonical copies are written that way.

One nit, not worth another round trip: the two polyline lines read `style={{strokeLinecap: 'round'} }` with a stray space before the closing brace, where the rest of the file uses `}}`.

Thanks Doug.

PR #464, opened 2026-09-17
feat: optimize richtoolbar

dcoffin88:
## Summary

Improves the rich text compose toolbar by grouping typography controls into a single dropdown menu. This reduces toolbar clutter and allows it to fit on a single row while keeping font family, font size, text color, and background color controls easily accessible.

## Changes

- Combined font family, font size, text color, and background color controls into a typography dropdown.
- Preserved the original text color and background color palettes.

**Old**
image

**New**
image

**New with Dropdown Open**
image

## Testing

Tested on my dev environment.

---


maathimself (approved):
Approving. The toolbar genuinely is cleaner on one row, and `ColorMenuSection` removes a lot of duplication between the text and highlight pickers.

Checked: the two color menus keep their original palettes, the custom-color input is still driven by a hidden `type="color"` and the `onMouseDown` + `preventDefault` pattern preserves the selection while the menu is open, which is the part that is easy to lose in a refactor like this.

I pushed one commit to the branch, since GitHub let me: `compose.toolbar.typography` was missing from `ptBR`, which landed in #471 after you opened this. Nothing to do with your change, just merge ordering. "Tipografia" is spelled the same in Italian so the uniqueness allowlist gained that pair too.

One thing I am taking on rather than asking you to: the color menu has three new hardcoded English strings, `Custom color...`, `No color` and `Remove color`. With ten locales now they would show in English for everyone else, and the i18n suite cannot catch them because it validates the locale files rather than the JSX. I will add the keys in a follow-up rather than hold this up.

Thanks for tightening this up.

Issue #441, opened 2026-09-14
create binaries for releases.


fahadalisarwar1:
I have to build desktop app with each update. i would like a automated flow where once you push the code, a binary is created automatically with tag for desktop/android apps. this will allow me to simply download the new app without cloning code and building binary again.

it seems that after updating server version, the desktop app stops responding and i see a blank page.

Thanks,

dcoffin88:
I’m the contributor who put together the native app packages.

My fork is still set up to monitor upstream releases and automatically build the native Desktop and Android packages whenever a new release is published. I use these builds for my own personal use, but you can also download the latest pre-built binaries directly from the Releases page without having to build them yourself. The built-in update flow also works with these releases.

https://github.com/dcoffin88/mailflow/releases

Please note that these builds are only self-signed, so your operating system may display a security warning during installation. Since they are built and distributed using self-signed certificates, use them at your own discretion.

maathimself:
Thanks @dcoffin88, and sorry for sitting on this. You offered to upstream it and I backburnered it, which is why your fork ended up carrying the load.

#467 builds all platforms on release and attaches the artifacts here, so the binaries come from this repo and the in-app updater points at these releases. It is largely your pipeline with the fork plumbing removed, plus your `.desktop` name fix. Android signing is configured now, so the APK ships alongside the desktop builds.

Two things I would value your eyes on, since you have actually run this and I have not:

1. macOS is built unsigned. I had excluded it on the wrong belief that Gatekeeper refuses unsigned apps outright rather than prompting. Your builds are the counterexample.
2. Unsigned installs now decline to auto-update on macOS as well as Windows, and link to the release instead. The darwin path runs `spctl --assess`, which an unsigned app cannot pass, so auto-update would otherwise download and silently discard every update.

@fahadalisarwar1 binaries will be attached to releases here from the next one onward.

dcoffin88:
Edit: Sorry, I responded before looking. I see now that you already covered my response for number 2 in PR #467 

@maathimself, correct, macOS makes it seem like it is refusing the install, but the user can still allow the app through Gatekeeper.

If you'd like, you could use wording similar to what I included in the original PR #124 README.md (below). That should cover the Gatekeeper steps for end users.

For number 2 we could simplify the update flow to just notify the user that an update is available with a Download Update button, which opens the latest release GitHub page; that way the end user is in full control of the download and install.

[...]

PR #323, opened 2026-07-29
fix: harden native app security boundaries

salmonumbrella:
## Summary

Hardens the Electron and Android wrappers before the first signed native release. This addresses the contributor-owned items in #301; CI least-privilege item 7 remains maintainer-owned as requested in the issue.

## Changes

- verify Electron release digests when GitHub provides them, require the downloaded Windows/macOS artifact to match the installed publisher/team, and hand Linux packages to the system installer without unattended `sudo`
- lock Electron top-frame navigation to the configured origin while allowing a bounded OIDC redirect chain, and require HTTPS except for warned localhost/RFC 1918 connections
- authenticate Android notification/update intents with a per-install secret before privileged dispatch
- replace the all-frame Android JavaScript interface and wildcard navigation rule with an exact-origin, main-frame-only message listener and exact origin matching
- restrict Android cleartext host configuration, redirect update downloads only across HTTPS, verify release digests when available, and scope FileProvider access to the update directory
- disable Android backups and third-party cookies, enforce renderer message origins, remove redundant wildcard message dispatch, and bound processed native action IDs

## Testing

- `frontend: npm test` — 1,413 passed
- `frontend: npm run lint`
- `frontend: npm run build`
- `backend: npx --yes node@22 node_modules/vitest/vitest.mjs run` — 706 passed
- `backend: npm run lint`
- `android: npm run android:sync`
- `android: ./gradlew testDebugUnitTest assembleDebug lintDebug`
- `electron: npm run electron:dist -- --linux dir`
- confirmed the committed Gradle 8.14.3 wrapper JAR SHA-256 against the official checksum
- not run: signed-artifact verification on Windows/macOS or native flows on physical devices

---

### Contributor License Agreement

By submitting this pull request I confirm that:

- [x] I have read and agree to the [Contributor License Agreement](../CLA.md).
- [x] My contribution is my own original work (or I have identified any
      third-party material and confirmed it is compatible with the CLA).
- [x] I have the right to submit this contribution under the terms of the CLA.

PR #467, opened 2026-09-17
feat(apps): publish native binaries for every platform

maathimself:
## Summary

Builds the desktop and Android apps when a release is published and attaches them to it, so users stop hand-building every version (#441).

Signing is optional on every platform. An unsigned build is published rather than withheld: the cost is an install-time warning, not an unusable download.

## Why macOS is included now

The previous version of this workflow excluded macOS from the build matrix, on the stated grounds that "an unsigned macOS app is refused by Gatekeeper rather than merely warned about". That is wrong. Gatekeeper prompts, with an Open Anyway path, and @dcoffin88 has been shipping unsigned `.dmg` builds from his fork that people install. So macOS now builds either way, like Windows.

## The part that needed care

Unsigned installs decline to auto-update, because `verifyPlatformSignature` cannot pin a download to a publisher the install does not have. That already held for Windows. It had to be extended to macOS: the darwin branch runs `spctl --assess`, which an unsigned app fails by definition, so without the extension an unsigned Mac install would download every update, fail verification and silently discard it, offering the user no link. Linux is unaffected, since those updates are gated by the SHA-256 digest from the release metadata.

## Other fixes in here

- **Trigger on `release: published`, not the tag push.** Triggering on the tag raced the release being created and would have left an empty release behind, breaking the existing `gh release create` step of the release process.
- **Skip Android when no keystore is configured, rather than failing.** The release job needed it, so a missing keystore blocked the desktop artifacts too.
- **Write `APPLE_API_KEY` to a file and export its path.** `@electron/notarize` documents the value as a file system path to the `.p8`, not its contents, so notarization could not have worked as written.
- **Declare the `workflow_dispatch` `version` input** that three steps already read. Manual runs fell back to the branch name and always failed.
- **Add a concurrency group** so a re-run supersedes an in-flight build instead of racing it.
- **Reject the literal `not set` team identifier** that `codesign` prints for an unsigned binary, so two unsigned builds cannot compare equal and pass as the same publisher.
- **Sync the Linux `.desktop` filename to the app id**, from the fix @dcoffin88 has been carrying in his fork.

## Testing

- `npm run lint` clean, `npm test` passes (2040).
- New case in `security.test.cjs` covers the `not set` team identifier. It fails against the previous implementation and passes against this one.
- The workflow parses as YAML, and every referenced script, output directory and action version was checked to exist.

Not yet exercised end to end: this cannot be until it runs against a real release.

## Before this can fully work

Android stays skipped until a keystore is configured (`MAILFLOW_ANDROID_KEYSTORE_BASE64`, `MAILFLOW_ANDROID_STORE_PASSWORD`, `MAILFLOW_ANDROID_KEY_ALIAS`, `MAILFLOW_ANDROID_KEY_PASSWORD`). Desktop artifacts publish without it.


Monkey7539 (commented):
I went through this at `a0dbff5`. Nothing below affects anyone yet, since no release has had assets attached, but four of them would bite on the first release this publishes.

## Worth fixing before merge

**1. `readWindowsSignature` fails for every file, so signed Windows installs would never auto-update.**
[`main.cjs#L839-L859`](https://github.com/maathimself/mailflow/blob/a0dbff5c6e2d9cfc9331148a07a118e6dd1f87db/frontend/packages/electron/main.cjs#L839-L859) passes the path as an argument after `-Command <script>`, but Windows PowerShell 5.1 treats everything after `-Command` as part of the command ([about_PowerShell_exe](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_powershell_exe?view=powershell-5.1)). So `$args[0]` is `$null`.

I ran the function as written on Windows 11 against `node.exe`, whose `(Get-AuthenticodeSignature).Status` is `Valid`. It throws `Cannot bind argument to parameter 'LiteralPath' because it is null`, and the path ends up glued to the end of the script (`ConvertTo-Json -Compress C:\Program Files\nodejs\node.exe`).

The bug dates to #323 and already broke `verifyPlatformSignature`. Now `canAutoInstallUpdates` swallows the error and returns `false`, so once `MAILFLOW_WINDOWS_CSC_LINK` is set every Windows install still takes the manual path, with nothing logged.

Passing the path out of band works: `Get-AuthenticodeSignature -LiteralPath $env:MAILFLOW_SIGNATURE_PATH`, with the path in `execFileSync`'s `env`, returned `{"status":"Valid","subject":"CN=OpenJS Foundation, ..."}` for the same file. Logging the error in that `catch` would have made this visible.

**2. The unsigned macOS build shows "damaged" on Apple Silicon, not an Open Anyway prompt.**
With `-c.mac.identity=null` ([L221](https://github.com/maathimself/mailflow/blob/a0dbff5c6e2d9cfc9331148a07a118e6dd1f87db/.github/workflows/publish-apps.yml#L221)), electron-builder 26.15.3 goes to `handleNullIdentity()` (`app-builder-lib/out/mac/MacTargetHelper.js`). That only logs "skipped macOS code signing". It doesn't apply an ad-hoc signature, and [its docs](https://www.electron.build/docs/features/code-signing/code-signing-mac/) say so.

The app then keeps Electron's prebuilt linker signatures, which don't seal the bundle (electron/electron#38694). Quarantined arm64 builds like that report "is damaged and can't be opened", with no Open Anyway in Privacy & Security. Intel gets the milder "cannot be verified" prompt (electron/electron#35524, electron-userland/electron-builder#8191).

@dcoffin88's `Universal.dmg` is built the same way. His CI log for run 35160138912 shows `skipped macOS code signing reason=identity explicitly is set to null`. So it isn't evidence of the Open Anyway path on current Macs.

Signing ad hoc when there is no certificate should give the milder prompt: `-c.mac.identity=-` resolves to `Identity("-")`, and the default entitlements template already has the JIT and library-validation keys. I couldn't test on a Mac, so this needs one browser-downloaded run on Apple Silicon before relying on it.

**3. Unsigned installs are never shown a link to the release.**
The workflow header, both warnings and the `canAutoInstallUpdates` comment all say these installs "link to the release instead", but nothing does:
- **Startup check:** `checkForUpdates(false)` reaches `notifyUpdateAvailable(false, { autoDownload: false })`, which sends `{ type: 'available' }` and returns at `if (!verbose) return;` ([L702-L722](https://github.com/maathimself/mailflow/blob/a0dbff5c6e2d9cfc9331148a07a118e6dd1f87db/frontend/packages/electron/main.cjs#L702-L722)). The only renderer consumer ignores anything but `downloaded` ([`ElectronNotificationBridge.jsx#L84`](https://github.com/maathimself/mailflow/blob/a0dbff5c6e2d9cfc9331148a07a118e6dd1f87db/frontend/src/components/ElectronNotificationBridge.jsx#L84)), so startup is silent.
- **Help > Check for Updates:** shows "A new version of MailFlow is available to download." with no action.

`updateInfo` never stores `release.html_url`, and `openDownload` only reveals an already downloaded file. Before this change those installs at least got a visible (if misleading) "could not be verified" toast.

It needs the release URL stored, a notification once per new version even when `verbose` is false, and an action that opens the release page.

**4. A manual run builds the dispatch branch, not the tag, and uploads it to that release.**
Both `actions/checkout@v6` steps have no `ref` ([L86](https://github.com/maathimself/mailflow/blob/a0dbff5c6e2d9cfc9331148a07a118e6dd1f87db/.github/workflows/publish-apps.yml#L86), [L264](https://github.com/maathimself/mailflow/blob/a0dbff5c6e2d9cfc9331148a07a118e6dd1f87db/.github/workflows/publish-apps.yml#L264)). For `workflow_dispatch` that is the branch selected in "Use workflow from". So backfilling `v3.5.1` from `main` would build `main` HEAD, stamp it 3.5.1, and upload it over that release's assets (`overwrite_files` defaults to true).

A mistyped version is worse. `softprops/action-gh-release` creates the missing release with a new tag from the default branch and publishes it as Latest (`make_latest` unset). Every desktop install polls `releases/latest`.

Two fixes would cover both:
- Check out `refs/tags/${{ github.event.release.tag_name || inputs.version }}`.
- Have `setup` run `gh release view "$VERSION"` and fail when it doesn't exist.

## Smaller

- **The release is Latest before its assets exist.** On `release: published`, startup checks during the roughly 10-minute build hit `notifyUpdateError('... no installer was found for this platform.')` ([L1020](https://github.com/maathimself/mailflow/blob/a0dbff5c6e2d9cfc9331148a07a118e6dd1f87db/frontend/packages/electron/main.cjs#L1020)), which toasts even for a non-verbose check. The release job needs every matrix leg plus Android to succeed ([L355-L358](https://github.com/maathimself/mailflow/blob/a0dbff5c6e2d9cfc9331148a07a118e6dd1f87db/.github/workflows/publish-apps.yml#L355-L358)), so one failed leg keeps every platform's files off an already published release. Two options: attach whatever built (`!cancelled()`, plus a final step that fails the run if a leg failed), or upload to a draft and publish at the end.
- **`always()` ignores cancellation.** Per the [workflow cancellation reference](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-cancellation), a running job with `if: always()` keeps going. So a superseding run in the concurrency group can still race the upload that the L30-31 comment says it prevents. `!cancelled()` keeps the skipped-Android behaviour.
- **Signed macOS installs can't auto-update either.** `verifyPlatformSignature` runs `codesign --verify` and `spctl --assess` on the downloaded `.dmg` ([L871-L896](https://github.com/maathimself/mailflow/blob/a0dbff5c6e2d9cfc9331148a07a118e6dd1f87db/frontend/packages/electron/main.cjs#L871-L896)), and electron-builder doesn't sign DMGs by default (`DmgOptions.sign` defaults to `false`). This predates the PR, but the new gate sends signed Macs straight into it. A rejected download is also left in Downloads under a new timestamped name on every launch. Options: verify the `.app` inside the mounted DMG, sign the DMG, or return `false` for darwin until one of those is tested.
- **A partial notarization config fails the build.** If only some of `APPLE_API_KEY` / `APPLE_API_KEY_ID` / `APPLE_API_ISSUER` are set, the stage step logs "signed but NOT notarized". `MacTargetHelper.getNotarizeOptions` then throws, because ID and issuer are always exported ([L213-L214](https://github.com/maathimself/mailflow/blob/a0dbff5c6e2d9cfc9331148a07a118e6dd1f87db/.github/workflows/publish-apps.yml#L213-L214)). Failing is fine, but it should fail in the resolve step, the way the certificate pair does.
- **A 1.2 s main-process block once #1 is fixed.** A working `readWindowsSignature` took about 1.2 s here, and it runs synchronously, with no timeout, on every check that finds an update. The installed exe's signature can't change while it runs, so caching the result per process avoids it.
- **Android `versionCode` comes from `github.run_number`** ([L296](https://github.com/maathimself/mailflow/blob/a0dbff5c6e2d9cfc9331148a07a118e6dd1f87db/.github/workflows/publish-apps.yml#L296)). A manual rebuild of an older release gets a higher code than newer releases, and Android then refuses those as downgrades.
- **Tests.** `canAutoInstallUpdates` has none. The new `macTeamIdentifier` test fails against `main` mostly because the function wasn't exported. The `not set` comparison it guards is still worth keeping.

## Checked and fine

- **Linux:** the `.desktop` output is unchanged. LinuxTargetHelper gives `/usr/share/applications/MailFlow.desktop`, `StartupWMClass=MailFlow`, before and after. `main.cjs` still sets `MailFlow.desktop`, so WM_CLASS / app_id and the `mailto` registration are unaffected. `linux.syncDesktopName` passes 26.15.3's `validateConfiguration`; electron-builder 27 drops the key.
- **Android:** the keystore secret names match between `setup`, the Android job and `build.gradle`. A skipped Android job doesn't skip the release job, and the missing `*.apk`/`*.aab` globs only warn.
- **Pinned actions:** the SHAs match `softprops/action-gh-release` v2.6.2 and `android-actions/setup-android` v3.2.2.
- **`security.test.cjs`** passes (9/9).

.github/workflows/publish-apps.yml
```diff
@@ -140,7 +274,7 @@ jobs:
           distribution: temurin
           java-version: "21"
 
-      - uses: android-actions/setup-android@v3
+      - uses: android-actions/setup-android@9fc6c4e9069bf8d3d10b2204b1fb8f6ef7065407 # v3
```
dcoffin88, line 324: `setup-android` now requires a package to be declared; without it, it will fail everytime.

```suggestion
      - uses: android-actions/setup-android@9fc6c4e9069bf8d3d10b2204b1fb8f6ef7065407 # v3
        with:
          packages: 'platform-tools'
```

frontend/package.json
dcoffin88, line 76: Per @Monkey7539 ad-hoc is the way to publish for macOS for the current iteration.

```suggestion
      "artifactName": "${productName}-${version}-Universal.${ext}",
      "identity": "-",
      "hardenedRuntime": false,
```

frontend/packages/electron/main.cjs
dcoffin88, line 897: Per @Monkey7539 

```suggestion
function readWindowsSignature(filePath) {
  const script = [
    '$signature = Get-AuthenticodeSignature -LiteralPath $env:MAILFLOW_SIGNATURE_PATH',
    '[pscustomobject]@{',
    '  status = [string]$signature.Status',
    '  subject = [string]$signature.SignerCertificate.Subject',
    '} | ConvertTo-Json -Compress',
  ].join('\n');
  const output = execFileSync('powershell.exe', [
    '-NoLogo',
    '-NoProfile',
    '-NonInteractive',
    '-Command',
    script,
  ], {
    encoding: 'utf8',
    windowsHide: true,
    env: {
      ...process.env,
      MAILFLOW_SIGNATURE_PATH: filePath,
    },
  });
  return JSON.parse(output);
}
```

dcoffin88, line 49:
```suggestion
let pendingUpdateDownloadUrl = null;
let lastNotifiedManualUpdateVersion = null;
```

dcoffin88:
```suggestion
function showInAppNotification({ title = '', message = '', type = 'info', actionLabel = '', action = '', actionUrl = '', persistent = false }) {
```

dcoffin88:
```suggestion
  const payload = JSON.stringify({ title, message, type, actionLabel, action, actionUrl, persistent });
```

dcoffin88, line 520:
```suggestion
          } else if (notification.action === 'open-update-release' && notification.actionUrl) {
            window.open(notification.actionUrl, '_blank');
          }
```

dcoffin88, line 711:
```suggestion
      releaseDate: updateInfo.releaseDate,
      releaseUrl: updateInfo.releaseUrl,
```

dcoffin88, line 720:
```suggestion
  
  if (!autoDownload) {
    const releaseVersion = updateInfo?.releaseVersion;

    // Background checks should notify only once for each new release.
    // A user-requested "Check For Updates" should always show the result.
    if (verbose || releaseVersion !== lastNotifiedManualUpdateVersion) {
      showInAppNotification({
        title: 'Update Available',
        message: 'A new version of MailFlow is available.',
        type: 'info',
        actionLabel: 'View Release',
        action: 'open-update-release',
        actionUrl: updateInfo.releaseUrl,
        persistent: true,
      });

      lastNotifiedManualUpdateVersion = releaseVersion;
    }

    return;
  }
```

frontend/packages/electron/main.cjs
```diff
@@ -714,7 +715,9 @@ function notifyUpdateAvailable(verbose = true) {
 
   notifyUpdateStatus({
     title: 'Update Available',
-    message: 'MailFlow is downloading the newest version for you.',
+    message: autoDownload
+      ? 'MailFlow is downloading the newest version for you.'
+      : 'A new version of MailFlow is available to download.',
```
dcoffin88:
```suggestion
    message: 'MailFlow is downloading the newest version for you.',
```

frontend/packages/electron/main.cjs
```diff
@@ -899,6 +902,44 @@ async function verifyDownloadedUpdate(filePath) {
...
+    }
+
+    if (process.platform === 'darwin') {
+      // A real Developer ID signature carries a team identifier; an ad-hoc or unsigned
+      // build reports "not set", which macTeamIdentifier returns as null.
+      return Boolean(macTeamIdentifier(readMacSignatureDetails(process.execPath)));
+    }
+
+    return true;
+  } catch {
```
dcoffin88:
```suggestion
  } catch (error) {
    console.error('Could not determine whether this install supports automatic updates:', error);
```

frontend/packages/electron/main.cjs
dcoffin88, line 1068:
```suggestion
      releaseName: release.name || release.tag_name,
      releaseVersion: release.tag_name,
```

dcoffin88, line 1070:
```suggestion
      releaseDate: release.published_at,
      releaseUrl: release.html_url || null,
```

.github/workflows/publish-apps.yml
```diff
@@ -1,26 +1,86 @@
...
+  # keystore is ours to generate rather than to buy. Whether one is configured is resolved
+  # here because the secrets context is not available in a job-level `if`.
+  setup:
+    name: Resolve build scope
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.matrix.outputs.os }}
+      android: ${{ steps.matrix.outputs.android }}
+    steps:
+      - id: matrix
```
dcoffin88, line 80: Per @Monkey7539 comment 4

```suggestion
      - name: Verify release exists
        shell: bash
        env:
          GH_TOKEN: ${{ github.token }}
          GH_REPO: ${{ github.repository }}
          VERSION: ${{ github.event.release.tag_name || github.event.inputs.version }}
        run: |
          set -euo pipefail
          
          echo "Verifying GitHub release: $VERSION"
          
          if ! gh release view "$VERSION" --repo "$GH_REPO" >/dev/null; then
            echo "::error::GitHub release '$VERSION' does not exist in '$GH_REPO'. Create/publish the release before running this workflow."
            exit 1
          fi
          
          echo "Release '$VERSION' exists."
      - id: matrix
```

.github/workflows/publish-apps.yml
```diff
@@ -1,26 +1,86 @@
...
       fail-fast: false
       matrix:
-        os:
-          - windows-latest
-          - macos-latest
-          - ubuntu-latest
+        os: ${{ fromJSON(needs.setup.outputs.matrix) }}
 
     steps:
       - uses: actions/checkout@v6
```
dcoffin88, line 114: Per @Monkey7539 comment 4

```suggestion
      - uses: actions/checkout@v6
        with:
          ref: refs/tags/${{ github.event.release.tag_name || github.event.inputs.version }}
```

.github/workflows/publish-apps.yml
dcoffin88, line 308: Per @Monkey7539 comment 4

```suggestion
      - uses: actions/checkout@v6
        with:
          ref: refs/tags/${{ github.event.release.tag_name || github.event.inputs.version }}
```

dcoffin88:
@Monkey7539, thanks for the analysis. The comment suggestions above should resolve your comments 1-4, and I have tested that the build was successful on my fork, on run 35282721241.

@maathimself, while testing the above changes across beta releases, I encountered a new intermittent issue with action-gh-release. In some runs, all eight asset uploads started successfully, but the action then failed with Error creating asset temp dir. The same upload process completed successfully on beta8, so this appears to be intermittent and may be worth monitoring.

maathimself:
Thank you both, this was a great review. Rebased onto 3.5.2 and pushed as 1b39623.

**On macOS, I tested your fix and it works.** I downloaded @dcoffin88's `v3.5.1-beta8` Universal dmg through a browser, so genuinely quarantined, onto an Apple Silicon M1. That build comes from his `test-native-binaries-pipeline` branch, which carries `identity: "-"` and `hardenedRuntime: false`, so it is the ad-hoc configuration you proposed. The dmg opened, copying to Applications worked, the first launch was refused, and allowing it once under System Settings > Privacy & Security ran it normally thereafter. Exactly the milder prompt you predicted.

So ad-hoc stays. I have scoped both flags to the build step rather than `package.json`, @dcoffin88, so that a future Developer ID certificate still gets its real identity and the hardened runtime notarization requires; in the shared `mac` block they would apply to the signed path too. Shout if you disagree.

I have not tested `identity=null` on Apple Silicon, so your "damaged" report stands as the only evidence either way for that path.

**Applied**

- **`readWindowsSignature`** now passes the path through the environment. Thank you for actually running it on Windows 11, @Monkey7539; that has been silently disabling auto-update for every signed Windows install since #323, and the `catch` now logs, since silence is what hid it.
- **Both checkouts take `refs/tags/<version>`**, and `setup` fails when the named release does not exist. The mistyped-version case worried me most: a new release published as Latest, which every desktop install polls.
- **`setup-android` declares `packages`**, which I would not have found.
- **`always()` is now `!cancelled()`.** The concurrency comment claimed a protection the condition then undid.

**One correction, and it is the same bug you found**

The "link to the release" suggestion could not have worked, and I applied it before noticing. `showInAppNotification` returns early when `window.__mailflowNativeBridgeReady` is set, which `ElectronNotificationBridge` sets on mount, so the injected toast does nothing in a running app; that helper is the fallback for when React is not up. And the bridge ignores every status except `downloaded`, so nothing handled it there either. The claim would have stayed false by a different route.

The payload now carries `canAutoInstall`, and the bridge renders a persistent notification with a **View Release** action when it is false. I traced that action rather than assume it: `window.open` hits `setWindowOpenHandler`, which passes `https:` to `shell.openExternal`.

**Deliberately not done**, all from @Monkey7539's smaller list, and better as separate issues than as more of this PR:

1. The release is Latest before its assets exist, so a check during the build toasts "no installer was found".
2. Signed macOS installs cannot auto-update either, since electron-builder does not sign DMGs by default.
3. A partial notarization config fails during the build rather than in the resolve step.
4. Now that `readWindowsSignature` works, it blocks the main process for about 1.2s per check and should be cached per process.
5. Android `versionCode` from `run_number` breaks ordering on a rebuild of an older release.
6. `canAutoInstallUpdates` has no tests.

@dcoffin88 on the intermittent `Error creating asset temp dir`: noted, and I would rather watch it than work around it. If it recurs on a real release we will have the run to look at.

PR #469, opened 2026-09-17
refactor(attachments): share one tiered extension list with antispam

Monkey7539:
## Summary

The shared attachment extension list agreed in #457: one canonical copy of the block / warn / notice / decoy tiers in `backend/src`, a bundled copy in the frontend, and a test that fails CI when they diverge. No scoring, verdict or warning changes.

## Changes

- **`backend/src/services/attachmentExtensions.js`** (new, canonical) exports `BLOCK`, `WARN`, `NOTICE` and `DECOY` with the frontend's current membership. The header says what each tier means.
- **`frontend/src/utils/attachmentExtensions.js`** (new) is the bundled copy, holding only the four sets, and `attachmentRisk.js` now imports it. The frontend image flattens `frontend/` into `/app`, so it cannot import from `backend/src`.
- **`backend/src/services/attachmentExtensions.test.js`** (new) fails CI when the two copies differ.
  - **Diff output:** a failure names the extensions missing on each side, for example `onlyInFrontend: ["hlp"]`.
  - **Tier overlap:** each extension is in at most one of block, warn and notice.
  - **Format:** every entry is a bare, lowercase extension.
  - **Double-extension rule:** fires for every decoy.
  - **No frontend packages:** the backend CI job loads the frontend copy without `frontend/node_modules`. That is why the sets live in their own import-free file rather than in `attachmentRisk.js`, which can then gain imports freely.
- **`ATTACHMENT_DOUBLE_EXT`** reads `DECOY` in place of `spamRules.js`'s private `PRESENTATION_EXTENSIONS`, which held the same 21 types.

**What this deliberately doesn't do.** `ATTACHMENT_EXECUTABLE` and the `attachment_is_executable` flag feature still read `EXECUTABLE_EXTENSIONS`, so `xlsm` and the other macro types still score 0.6 for now. Moving the rule to `BLOCK` changes what it scores, and #457 agreed that belongs with the weight re-tune and the `bodyStructure` attachment list at insert. When that lands, the switch changes the rule's membership as follows:

- **Leaves the rule:** `docm xlsm pptm dotm xlsb xlam` (to warn), plus `class bash zsh command rb` (dropped).
- **Joins the rule:** 37 types, the 21 discussed in #457 plus the 16 added in #462: `msp mst reg lnk iso img vhd vhdx dll application appx run deb rpm chm inf scf url ade adp ws msc xll py pyw pyz pyzw appref-ms msu diagcab sct wsc settingcontent-ms search-ms library-ms website rdp`.
- **Tests that move with it:** the `EXECUTABLE_EXTENSIONS` test in `spamTokenizer.test.js` and the flag feature.

A comment on `EXECUTABLE_EXTENSIONS` points at the pending switch.

## Testing

- **Drift in both directions fails the parity test:**
  - `hlp` added to the frontend `BLOCK` only fails with `onlyInFrontend: ["hlp"]`;
  - `rtf` removed from the backend `DECOY` only fails on the `DECOY` set.
- **No behaviour change:**
  - `DECOY` is set-equal to the old `PRESENTATION_EXTENSIONS`, and nothing else read that set.
  - Every frontend tier is set-equal to its previous contents.
  - The `spamTokenizer.js` change is a comment only.
- **Backend:** `npm run lint` and `npm run lint:plugins` clean, `node --check src/index.js` passes, `npm test` passes (1621/1621).
- **Frontend:** `npm run lint` clean, `npm test` passes (2039/2039), `npm run build` succeeds.

---


maathimself (approved):
Verified, and the drift test does what it claims.

I mutated the two copies in both directions: adding `hlp` to the frontend `NOTICE` only fails with `onlyInFrontend: ["hlp"]`, and removing `rtf` from the backend `DECOY` only fails on that set. The failure names the offending extension rather than just asserting inequality, which is what makes the guard usable when it fires months from now.

Also confirmed behavior is unchanged: `DECOY` is set-equal to the `PRESENTATION_EXTENSIONS` it replaces, and the double-extension and download-warning cases from #460 through #462 all classify identically after the merge.

Leaving `ATTACHMENT_EXECUTABLE` on `EXECUTABLE_EXTENSIONS` is the right call. Moving it changes what it scores, and that belongs with the weight re-tune and the `bodyStructure` attachment list, as agreed. The comment pointing at the pending switch, and the list of exactly which types join and leave the rule, will make that change much easier to review when it comes.

One note for whoever picks up the re-tune: until it lands, `xlsm` and the other macro types still score 0.6, so the accountant's spreadsheet from #457 is unchanged for now. That is expected, not an oversight, but it means this PR is groundwork rather than the fix.

Thanks for turning this around so quickly.

PR #470, opened 2026-09-17
refactor(attachments): tidy the risky-download confirm from #459

Monkey7539:
## Summary

The two follow-ups noted on #459 (in #457):

- **Separator:** the " — " before "click again to download anyway" sits outside the translated string, so no locale can change that punctuation.
- **Sentinel:** `DOWNLOAD_ALL = 'all'` relies on an attachment part never being that literal, where a `Symbol` makes it structural.

## Changes

- **`message.attachmentRisk.confirm` is replaced by `message.attachmentRisk.armed`**, `"{{label}} — click again to download anyway"`. The string wraps the text it follows: the attachment's own warning, or "Download all". Each of the 9 locales keeps its existing wording, with the separator now inside its own string.
- **`DOWNLOAD_ALL` is now `Symbol('downloadAll')`.** It is only ever a `useState` value compared with `===`, so nothing else changes.
- **New placeholder check in `i18n.test.js`.** Every locale must use the same `{{...}}` names as `en.json` for the keys they share. With the warning now carried in `{{label}}`, a translation that dropped the placeholder would hide it exactly when the user is asked to confirm. All locales already pass, so no allowlist is needed.

Rendered text is unchanged in all 9 locales: with real i18next and `escapeValue: false`, the armed text is identical to the previous concatenation for Download all and for block, warn and double-extension warnings.

## Testing

- **Render test stub:** the `react-i18next` stub in `MessagePane.render.test.js` now shows a `{{label}}` value in brackets, so the tests can check which text is wrapped. No other `t()` call passes `label`.
- **New render tests:**
  - A risky file's own confirm wraps its warning (`message.attachmentRisk.armed[message.attachmentRisk.doubleExt]`), arms nothing else, and downloads nothing on the first click.
  - An attachment whose part is literally `"all"` does not arm Download all. Changing the `Symbol` back to `'all'` fails exactly this test.
- **Download recording:** the test's fetch mock now also records single-attachment downloads, so "downloads nothing" is a real check. A mutant that starts the download on the first click fails it.
- **Dropped placeholder:** removing `{{label}}` from the Polish string fails the new i18n check with `pl message.attachmentRisk.armed: {{}} but en has {{label}}`.
- `npm run lint` clean, `npm test` passes (2042/2042, including i18n 1458/1458), `npm run build` succeeds.

---


maathimself (approved):
Verified, and both mutation checks bite.

Reverting `DOWNLOAD_ALL` to the `'all'` string fails the test for an attachment whose part is literally that, and removing `{{label}}` from the Polish string fails the new i18n check with exactly the message you quoted: `pl message.attachmentRisk.armed: {{}} but en has {{label}}`. That second one is the useful part: a translator dropping the placeholder would silently hide the warning at the precise moment someone is being asked to confirm a risky download, and nothing else would have caught it.

Confirmed all nine locales keep their existing wording with the separator moved inside the string, and that the rendered text is unchanged. Frontend suite 2042 passing, lint clean.

Making the render-test stub show `{{label}}` in brackets so the tests can assert which text is wrapped was a neat way to verify this without pulling in real i18next.

Keeping the em dash as the separator, since it reads correctly in the UI.

Thanks for picking these up, they were genuinely minor and easy to leave undone.

PR #393, opened 2026-08-21
feat: add Czech translation

sympatickejsam:
## Summary

Adds complete Czech localization to MailFlow, making the interface accessible to Czech-speaking users.

The translation was prepared with AI assistance. I am a native Czech speaker and will address any wording issues identified during review.

Closes #385.

## Changes

* Added the complete Czech locale in `frontend/src/locales/cs.json`
* Registered Czech in the localization system
* Added Čeština to the language selector and settings search
* Added Czech plural forms and pluralization tests
* Whitelisted intentional identical technical and brand terms
* Added Czech strings introduced through v3.2.4, including Mailbox Cleanup and asynchronous folder-emptying states
* Synced the branch with the current `main`

## Testing

* Frontend: `npm test` — 1,639 tests passed
* Frontend: `npm run lint`
* Frontend: `npm run build`
* Backend: `npm test` — 1,172 tests passed
* Backend: `npm run lint`
* Backend: `npm run lint:plugins`

---

### Contributor License Agreement

By submitting this pull request I confirm that:

* [x] I have read and agree to the [Contributor License Agreement](../CLA.md).
* [x] My contribution is my own original work (or I have identified any
  third-party material and confirmed it is compatible with the CLA).
* [x] I have the right to submit this contribution under the terms of the CLA.

Issue #255, opened 2026-07-13
[Feature Request] Allow customizing prompt or auto-adapting language for AI email summary


Maomaoxion:
### Describe the feature request
Currently, the built-in AI email summary feature (within the AI operations) does not expose its underlying prompt for customization. When using this feature directly, it always defaults to outputting the summary in English.

For non-English speaking users, it would be much more convenient if we could control the output language.

### Proposed Solution
I would like to propose two potential solutions:

1. **Customizable Prompt for AI Summary (Preferred):**
   Expose the prompt template for the built-in AI summary operation—similar to how the **Custom AI Operations** currently work. This would allow users to manually edit the prompt and specify their preferred output language (e.g., adding "Response in Chinese").

2. **Auto-Adapt based on UI Language:**
   Alternatively, the summary feature could automatically detect the user's current interface/system language setting in Mailflow and dynamically request the AI to output the summary in that language.

### Additional Context
Allowing prompt customization for the default summary tool would keep the configuration experience highly consistent with the existing "Custom AI Operations" settings while providing maximum flexibility.

computrav:
**Alternative Implementation for this (since users can now add their own AI Actions)**: Just make "Summarize" a pre-installed AI Action (and like other actions, show the prompt). this would also allow users to rename/modify/remove it if they wish.

**Regarding it working for different languages, here is the basic custom AI prompt that I use, which I call "Translate & Summarize":**
> Translate this email into clear, natural language, preserving meaning and tone. If the email is in English, translate it to German. Otherwise, translate it to English. Output in 2 sections: section 1 should be a bullet summary of the message, and section 2 should be the translation.

Maomaoxion:
@computrav Thanks for the suggestion! I completely agree that making "Summarize" a pre-installed AI Action (and exposing its prompt) is a very clean way to solve this.

Regarding the multi-language support: if the default "Summarize" remains hardcoded/separate and we have to create a new, custom "Translate & Summarize" action to handle different languages, it would result in redundant/duplicate actions in the UI, which isn't very elegant. 

If the default "Summarize" itself is simply a customizable pre-installed action, users can just directly modify its default prompt to handle translation or specify their preferred output language. This keeps the interface clean and avoids unnecessary duplication.

maathimself:
[...]

PR #471, opened 2026-09-18
feat: add Brazilian Portuguese (pt-BR) translation

prudenciooficial:
## Summary

Adds a complete Brazilian Portuguese (`ptBR`) translation of the UI, following the same pattern as the Czech locale (#393). All 1442 keys from `en.json` are translated; `{{placeholders}}` and `_one` / `_other` plural forms are preserved.

## Changes

- `frontend/src/locales/ptBR.json` — new locale file, key order mirrors `en.json`
- `frontend/src/i18n.js` — register the `ptBR` resource
- `frontend/src/components/AdminPanel.jsx` — add "Português (Brasil)" to the language selector and search keywords
- `frontend/src/aiActions.js` — built-in Summarize prompt responds in Brazilian Portuguese for this locale (same mechanism as the other locales, #255)
- `frontend/src/locales/i18n.test.js` — extend `SAME_VALUE_ALLOWED` for words Portuguese shares with Spanish/Italian (Responder, Cancelar, Enviar, Copiar, …) and for the brand/technical strings that are identical everywhere
- `README.md` — list Brazilian Portuguese (and Czech, which was missing) under Multi-language UI

Only the translation itself is new; no behaviour changes outside locale registration.

## Testing

- `npm run lint` in `frontend/` — clean
- `npm test` in `frontend/` — 2043 pass / 0 fail, including `i18n.test.js` (source coverage, key coverage, value uniqueness, placeholder parity)
- `vite build` — succeeds

---

🤖 Generated with the help of [Claude Code](https://claude.com/claude-code)


maathimself (approved):
Verified rather than taken on trust, and it holds up.

- **Key parity is exact**: 1442 keys in `en.json`, 1442 in `ptBR.json`, nothing missing and nothing extra.
- **Placeholders are intact** across every shared key, including the `_one` / `_other` plural forms.
- **Registration is complete** in all three places: the `i18n.js` import and resource, the `aiActions.js` summarize locale, and the AdminPanel selector.
- Merges clean against current main; lint, the full frontend suite and the build all pass.

The `i18n.test.js` change is the part I looked at hardest, since that file is what keeps the other locales honest. It only extends existing allowlist groups, and where Portuguese matches a different locale it adds a separate group rather than merging two, for example `'compose.bcc': [['de','en','it'], ['es','ptBR']]`. That keeps the constraint tight instead of loosening it, which is the right instinct.

I also went through the 46 strings left identical to English. All of them are brand names, technical placeholders or genuine loanwords: Gmail, iCloud, Todoist, Ko-fi, `imap.gmail.com`, `000000`, the CSS sample, and words like Layout, Status, Login and Social that are the same in Portuguese. Nothing looks skipped.

**One thing to be aware of, and it is not your fault.** #464 adds a `typography` key to the nine locales that existed when it was opened, so it does not include `ptBR`. Whichever of the two merges second leaves the other's locale short a key, and `key coverage — every key must appear in every locale` fails. I confirmed it by merging both. I will add the missing key as part of landing the second one, so there is nothing for you to do.

Thank you for following the Czech locale's pattern so closely, it made this quick to check.

